Issue #71, opened 2026-04-28
Add support for EcoFlow STREAM Ultra X


OpaWoffel:
### Device Type

New Device Support

### Problem or Use Case

Current behavior:
The integration (version 1.12.0) discovers my EcoFlow STREAM Ultra X device, and it shows up under "Devices & Services" in Home Assistant. However, no entities (sensors, switches, etc.) are created. The logs contain no errors related to ecoflow or stream.

### Proposed Solution

Expected behavior:
After adding the device, I expect to see entities for:

Battery state of charge (SOC) in %

Input power (solar and AC charging) in W

Output power (grid feed-in/backup) in W

Device status (idle/charging/discharging)

Temperature

The device uses the official EcoFlow public API (via [api-e.ecoflow.com](https://api-e.ecoflow.com/)) with valid access/secret keys. My goal is to monitor the battery level to automate charging/discharging based on electricity prices.

My device:

Model: EcoFlow STREAM Ultra X (3.84 kWh capacity, 800W grid-tied AC output)

Connection: Wi-Fi/Bluetooth

Integration: shuette42/ecoflow-energy-ha v1.12.0

shuette42:
The integration only registers entities for device families it has a parser for, and the STREAM series isn't in that list yet. That's why HA shows the device but no sensors, switches or numbers attach to it. It's not a config or auth problem on your side, just a missing parser.

Two things would help me scope what's involved:

1. **The SN prefix of your STREAM Ultra X.** Only the first 4 characters are needed, the rest can be redacted. The integration currently knows `HJ31` and `HJ32` (PowerOcean), `R351` and `R331` (Delta 2 Max), and `HW52` (Smart Plug). STREAM is almost certainly a new prefix I haven't seen yet, and knowing it lets me match the device against the protocol data I have. You'll find the SN on the sticker on the device (usually side or back), or in the EcoFlow mobile app under Device Info → Serial Number.

2. **Whether you'd be willing to run a test build once there's something to try.** In practice that means installing an alternative version of the integration via HACS as a custom repository, and if something looks off, downloading a diagnostics file (Settings → Devices & Services → EcoFlow Energy → ⋮ → Download diagnostics) plus debug logs. No compiling, no command line, just a couple of clicks and sending me a JSON file.

The first iteration would likely be read-only sensors only. Controls (switches, charge limits, mode changes) come later, if at all, and only once the read path is solid.

No rush, whenever you have the info post it here.

[...]

PR #97, opened 2026-06-11
feat: add Stream AC Pro support

Hendrixx-hue:
## Related Issue

Adresses #71 but for Stream AC Pro.

## Summary

Adds initial Enhanced Mode support for EcoFlow Stream AC Pro (BK31), including device detection, protobuf telemetry parsing, Stream sensor entities, and a Backup Reserve number control.

## Changes

- Added Stream AC Pro device type detection by product name and BK31 serial prefix.
- Added conservative Stream protobuf parser for observed telemetry fields.
- Added Stream sensor and number entity definitions with English/German translations.
- Added Enhanced Mode MQTT parsing and Backup Reserve SET support.
- Added tests for device routing, parser behavior, entities, and coordinator parsing.
- Updated README and CHANGELOG.

## Checklist

- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated

## Transparency

I used GPT for finding and deciphering protobuf.
I also cross checked with tolwi/hassio-ecoflow-cloud who has implemented already sensors in HA for BK31.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.15.0] - 2026-06-12
+
+### Added
+- Initial Stream AC Pro (BK31) support in Enhanced Mode, including device detection, protobuf telemetry parsing, Home Assistant sensor entities, and a Backup Reserve number control.
+
 ## [1.14.0] - 2026-05-18
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Initial Stream AC Pro (BK31) support in Enhanced Mode, including device detection, protobuf telemetry parsing, Home Assistant sensor entities, and a Backup Reserve number control.
+- Stream AC Pro Energy Dashboard sensors (solar, home, battery charge/discharge) derived via Riemann-sum integration from the live power telemetry.
 
 ## [1.14.0] - 2026-05-18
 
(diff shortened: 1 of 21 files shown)


shuette42:
Thanks for this, @Hendrixx-hue. Strong first contribution. The sources are well documented and the parser plus sensor path is solid and tested.

I reworked a few things before merging:

The Backup Reserve SET command was sending JSON, but this topic does not accept JSON writes, only the binary format it expects. That meant the write path would not have reached the device. Hard to catch without real hardware, so no fault on your end. I switched it to the binary format the topic expects.

I also added the Energy Dashboard sensors (solar, home, battery charge/discharge) and extended the test coverage, including regression tests for the write path. The suite is now at 944 green.

A beta is now available on the HACS beta channel as a pre-release (v1.15.0-beta.1). If you run HACS in beta mode you can pull it directly.

One honest caveat: I do not own a Stream AC Pro / BK31. The Backup Reserve write path is a well-founded hypothesis based on observed protocol behavior, but it has not been confirmed on real hardware. The sensors should be fine, the write path needs on-device verification.

If you can install the beta on a real device, please test carefully:

1. Do the sensors show plausible values after an HA restart (SoC, power, etc.)?
2. Move the Backup Reserve slider in HA, then check in the EcoFlow app whether the new value actually reaches the device.
3. After a few seconds, does the value hold, or does it snap back to the old value?
4. Do the new energy sensors (solar, home, battery charge/discharge) show up in the HA Energy Dashboard?

Steps 2 and 3 are the important ones, since that is the write path I cannot verify here. If you can run those on your BK31 and report back what you see, that would help a lot. The full release follows once the write path is confirmed on a real device.

Hendrixx-hue:
Thank you @shuette42 for merging the PR so quickly! I appreciate it.

I have tested the v1.15.0-beta.1 in my local home assistant from the HACS store:

1. Yes. They do show plausible values and the integration also added the sensors correctly after one reboot of HA.
2. No. The sliders are not working currently. I have read, that some more reverse engineering is needed. I am doing it now. At this moment, I would state these are read only.
3. They stay at the manual set value, no snap back or old values.
4. yes. The new energy sensors are working fine in the energy dashboard. 

I am currently working on a clean "read-only commit/fix" for the outlets (switch state and power), LED-brightness, backup reserve slider and some more fields I discovered for the BK31.

Hendrixx-hue:
Update: I found some more bugs while using the ac outlets, which resulted in wrong total power values.

I am working on a new commit.

PR #99, opened 2026-06-14
feat(stream): add AC Pro read-only telemetry

### LINKED ISSUE #98: feat: add support for Stream AC Pro
### Device Type

New Device Support

### Problem or Use Case

I want to integrate the Ecoflow Stream AC Pro (BK31) device.

### Proposed Solution

Since I own one of it, I can make some tests if needed. 

### Alternatives Considered

_No response_

### Additional Context

_No response_

Hendrixx-hue:
## Related Issue

Closes #98 

## Summary

Expose Stream AC Pro AC outlet state and power as read-only entities, add LED brightness diagnostics, and keep unconfirmed outlet/LED write paths out of the integration.

Add signed AC grid connection power for the app Netz-Anschluss value, preserve battery power as the dedicated battery path, and keep meter-dependent solar/home values disabled by default.

Remove raw Wh battery-energy entities, mark Ah capacity counters as disabled diagnostics, and document Stream AC Pro Energy Dashboard and upgrade cleanup guidance.

## Changes

- Add read-only Stream AC Pro AC outlet state and outlet power entities.
- Add LED brightness as a disabled diagnostic read-only sensor.
- Add signed AC grid connection power matching the EcoFlow app Netz-Anschluss value.
- Keep battery power derived from the dedicated battery telemetry path instead of grid connection fallback frames.
- Keep meter-dependent solar/home energy sensors disabled by default as diagnostics.
- Remove raw Wh battery energy entities in favor of kWh charge/discharge energy sensors.
- Mark Ah battery capacity counters as disabled diagnostic entities.
- Update README and CHANGELOG for Stream AC Pro behavior, Energy Dashboard setup, and upgrade cleanup notes.
- Extend Stream parser, entity, translation, and energy tests.

## Checklist

- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated

## Transparency 

Used GPT 5.5 for debugging protobuf and vibe coding changes.
Tested with
- Home Assistant OS
- Core 2026.6.3
- Supervisor 2026.06.1
- Operating System 17.3
- Frontend 20260527.6
- on a Raspberry Pi 4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.16.0] - 2026-06-14
+
+### Added
+- Stream AC Pro read-only telemetry for individual AC outlet state and power, LED brightness diagnostics, and signed AC grid connection power matching the EcoFlow app "Netz-Anschluss" value.
+
+### Changed
+- Stream AC Pro is now modeled as an AC-coupled battery: battery charge/discharge energy stays enabled for the Energy Dashboard, while meter-dependent solar/home values remain disabled diagnostics unless an EcoFlow-compatible meter is paired in the app.
+- Ah battery capacity counters are disabled diagnostic entities.
+
+### Removed
+- Removed unconfirmed experimental write paths for Stream AC Pro AC outlets and LED brightness.
+- Removed raw Wh battery-energy entities in favor of the kWh battery charge/discharge energy sensors.
+
 ## [1.15.0] - 2026-06-12
 
 ### Added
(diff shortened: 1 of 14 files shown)


shuette42:
Thanks for this, and for testing it on real hardware (HA OS, Core 2026.6.3, Pi 4). That makes a real difference for confidence in the merge.

I pushed one follow-up commit onto your branch. Here is what changed and why:

**LED brightness:** Now reads only the live field. The set/ack path could overwrite the live reading, which made the value flap. Reading the live field alone keeps it stable.

**Battery energy:** Dropped two raw battery-energy fields. No entity consumed them.

**Charge/discharge state:** Now derived centrally in the coordinator. The parser value was already discarded on the live path, so this just removes a dead branch.

**Tests and docs:** Added tests around battery-power field fallback and precedence, plus the single-frame outlet enable+mirror case. Also a README note that existing v1.15.0 installs keep solar, home, and grid power enabled after the diagnostic reclassification.

All 957 tests pass and CI is green. I will squash-merge this now. It ships in v1.16.0.

One possible next step if you are interested: write support for LED and outlet control. That needs the relevant SET frames captured from a device. If you can grab those, you would be well placed to drive that part.

After v1.16.0 is out, would you want to verify the final behavior on your device? No obligation, just an offer if it is convenient.

Issue #111, opened 2026-07-08
Delta 3 Max Plus (D3M... serial) never contacted in Enhanced Mode – zero MQTT attempts logged


NatsumiCH:
### Home Assistant Version

HA Core: 2026.7.1 (HAOS 18.1)

### Integration Version

ecoflow_energy integration: v1.15.0

### Device Type

Other / Multiple

### Connection Mode

Enhanced (MQTT Protobuf)

### Bug Description

Delta 3 Max Plus (serial prefix D3M) never connects in Enhanced Mode.

The Stream AC Pro (BK31FEBB2H950021) connects and streams normally via 
"WSS Enhanced" (mqtt_status.connected=true, data_receiving=true, ~2-3s 
update interval, 41 data keys). Entities are created correctly for it.

The Delta 3 Max Plus (D3M1ZE1A9J460241) and its paired Smart Meter 
(BK21Z1BB7H4U4050) never appear in the integration diagnostics 
`devices` list at all, and produce *zero* log lines even with the 
logger set to DEBUG for `custom_components.ecoflow_energy` — no 
connection attempt, no error, nothing. No entities/device are ever 
created for them in HA.

Question: Is the Delta 3 Max Plus product family currently unsupported 
by the Enhanced Mode WSS bridge (as opposed to e.g. Delta 2 Max)? If 
Standard Mode is meant to support it, what additional setup 
(Access/Secret Key from the Developer Portal, specific auth_method, 
etc.) is required, since simply switching modes broke the whole 
integration for me rather than adding the missing device?

### Steps to Reproduce

1. Bind a Delta 3 Max Plus + its paired Smart Meter to the EcoFlow 
   Energy integration alongside a Stream AC Pro, using Enhanced Mode.
2. Disable + re-enable the config entry (full integration reload).
3. Open the EcoFlow app and actively view/select the Delta 3 Max Plus 
   live screen while reloading, in case Enhanced Mode only streams the 
   most-recently-viewed device.
4. Check integration diagnostics and DEBUG logs for 
   custom_components.ecoflow_energy — no connection attempt for the 
   Delta 3 Max Plus or Smart Meter is ever logged.
5. (Optional/risky) Switch global mode to "Standard – Official EcoFlow 
   API" — this broke *all* devices, including the previously working 
   Stream AC Pro (no entities/data at all). Reverting to Enhanced 
   restored the Stream AC Pro immediately.

### Relevant Log Entries

```text
[...]

Issue #110, opened 2026-07-06
All entities show 'unknown' on DELTA 3 Max Plus despite API returning data (availability_stage: stale)


TongueJones:
### Home Assistant Version

2026.7.1

### Integration Version

1.14.0

### Device Type

Delta 2 Max

### Connection Mode

Standard (HTTP Polling)

### Bug Description

All 62 entities for the DELTA 3 Max Plus show `unknown` state in Home Assistant despite the integration successfully connecting to EcoFlow's API and receiving data.

## What's working

- MQTT is connected (`status: connected`)
- HTTP API polling is working — 27 data keys returned every 30 seconds (`http_ok` events visible in diagnostics event log)
- Device is online and visible in the EcoFlow app
- API credentials (developer portal access key + secret key) are correct
- Device was auto-discovered during integration setup

## What's wrong

Despite data arriving successfully, all entities remain `unknown`. The diagnostics show:

- `availability_stage: stale`
- `mqtt_status.data_receiving: false`
- `snapshot.captured: false` / `snapshot.source: none`

The integration appears to receive data but does not map it to entities or mark itself as available.

### Steps to Reproduce

1. Install EcoFlow Energy integration via HACS
2. Configure with developer API credentials (access key + secret key)
3. Device is auto-discovered
4. All entities show `unknown` indefinitely

### Relevant Log Entries

```text

```

### Diagnostics

```json
{
  "device_sn": "D3M1XXXXXXXXXXX",
  "device_name": "DELTA 3 Max Plus",
  "enhanced_mode": false,
  "availability_stage": "stale",
  "mqtt_status": {
    "status": "connected_stale",
    "connected": true,
    "data_receiving": false,
    "uptime_s": 1628.4,
    "reconnect_attempts": 0
  },
  "data_freshness": {
    "last_mqtt_age_s": 62.3,
    "update_interval": "0:00:30",
    "http_fallback_active": false
  },
  "snapshot": {
    "source": "http",
    "age_s": 2.3,
    "key_count": 27,
    "captured": true
  },
  "data_key_count": 27,
  "event_log": [
    { "type": "mqtt_connect", "detail": "TCP Standard" },
    { "type": "http_ok", "detail": "keys=27" },
    { "type": "mqtt_data", "detail": "keys=27" }
  ]
}
```

### Additional Context

Switching to Enhanced mode results in `availability_stage: unavailable` with zero data keys returned, making Standard mode the better option despite the stale issue.

[...]

PR #112, opened 2026-07-14
feat: add Delta 3 Max Plus support (Standard mode, read-only)

shuette42:
## Summary

Adds initial read-only support for the EcoFlow Delta 3 Max Plus (SN prefix D3M1) in Standard mode and makes unsupported devices visible instead of silently skipping them.

Ref #100
Ref #110
Ref #111

## Root cause of the linked issues

- Enhanced-mode entries (#100, #111): the app API returns an empty productName, and the D3M1 serial prefix was unknown, so the devices were skipped with only a DEBUG log - diagnostics showed `device_count: 2, devices: []` with no trace.
- Standard-mode entries (#110): the product name "DELTA 3 Max Plus" matched the generic "delta" keyword and the device got the Delta 2 Max parser, whose field map matches none of the Delta 3 quota keys - all entities stayed `unknown` with `availability_stage: stale`.

## Changes

- Device classification: "delta 3" keyword checked before "delta", D3M1 serial prefix mapped, re-classification on setup heals installs stored with the wrong type
- New `delta3_http.py` parser: 24 quota key mappings (direct scaling), charge/discharge state enum, output-flow states (on = value != 4), per-outlet AC power from the nested array
- 21 sensors + 7 binary sensors, English and German translations
- Unsupported devices now produce one setup warning each and a `skipped_devices` diagnostics section
- Delta 3 diagnostics include the raw quota snapshot (serial-redacted) so beta testers can confirm and extend the field map
- MQTT quota pushes accept both `param` and `params` wrappers; HTTP polling remains the primary data path

Deliberately deferred: SET commands (need hardware verification), energy (kWh) sensors, Enhanced-mode support for Delta 3, and the paired Smart Meter (BK21).

## Testing

- Full suite: 1019 passed (63 new tests: parser, classification, skip path, diagnostics redaction, MQTT routing, end-to-end integration)
- Field map researched from community integrations verified against live Delta 3 Max Plus devices; hardware validation by the issue reporters is the acceptance gate for the beta

## Status

Draft until the Docker validation run completes.


EdwardK9:
How does this get added? I have a delta 3 classic and have the same thing of every sensor just saying unknown.

PR #126, opened 2026-07-22
fix: reclassify unknown app device types

### LINKED ISSUE #118: Device type classification falls back to product name when API returns device_type: unknown
Devices whose serial prefix is not in the mapping table end up with no entities at all, and the log only says the device was skipped because no parser is available.

The classification reads `device_type` from the API response and only falls back to the serial-number lookup when that key is missing. For some devices the API returns `"unknown"` as the value, and for some it returns an empty product name alongside it. Since the key is present, the fallback never runs, and the device is discarded.

The fix is to treat `"unknown"`, an empty string and `None` as invalid values so the serial-number classification takes over.

Reported in #89 with the code paths identified there.

An external PR for this is welcome. It should cover all three invalid values and add one test per case. Prefix additions to the mapping table are a separate topic and belong in their own PR, not in this one.

rimraf:
## Related Issue

Closes #118

## Summary

Ensures devices are classified locally when the App API returns an invalid device type.

## Changes

- Treat `"unknown"`, an empty string, and `None` as invalid device types.
- Fall back to classification using the product name and serial-number prefix.
- Add a separate regression test for each invalid value.
- Add a changelog entry.

## Checklist

- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated
- [ ] Version bumped in `manifest.json` (if code change)
- [x] README updated (not applicable—no documentation changes required)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 - Ah battery capacity counters are disabled diagnostic entities. (beta.2)
 
 ### Fixed
+- App API devices reported with an unknown device type are now classified locally using their product name or supported serial-number prefix.
 - The Delta 2 Max AC Charge Speed slider now reads and writes the actually configured charging speed. It previously displayed the rated maximum (2400 W), never picked up changes made in the EcoFlow app, and every value written from Home Assistant effectively ended up at 400 W because the command hardcoded the governing parameter. (beta.7)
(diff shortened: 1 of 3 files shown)


shuette42:
Thanks for the fix, cleanly done including the three test cases (unknown/empty/null). I only merged main into the branch to resolve a CHANGELOG conflict from two beta releases in the meantime, the change itself went in unchanged. This closes #118 and ships with v1.15.0-beta.10.

PR #129, opened 2026-07-22
fix: recognize base DELTA 3 (P321 SN prefix) in Enhanced mode

EdwardK9:
## Related Issue
Not filed as a separate issue — found while getting a personal DELTA 3
Classic working, via a diagnostics download.

## Summary
DELTA 3 Classic devices are silently skipped as unsupported in Enhanced
mode because the app-auth device list reports them with an empty
productName, so the existing keyword-based classification never matches.

## Changes
- Add `P321` serial prefix to `_SN_PREFIX_MAP` in `const.py`, alongside
  the existing `D3M1` (Max Plus) entry
- Add a regression test confirming `P321` classifies as `delta3`

## Checklist
- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated
- [ ] Version bumped in `manifest.json` (if code change)
- [ ] README updated (if new sensors, features, or behavior changes)


shuette42:
Thanks for the fix and the diagnostics-based analysis. Two small adjustments from my side: I replaced the serial number in the test with a synthetic placeholder (we keep full serial numbers out of the repo, tests included) and merged main to resolve a CHANGELOG conflict. The fix ships with v1.15.0-beta.10.

One ask: after updating, please check whether the controls (switches, charge/discharge limits) also work correctly on the Delta 3 Classic. The sensors are confirmed by your diagnostics, but I could not verify the SET commands without the device.

EdwardK9:
Hi,
Sorry about that I totally missed it when I went through.
I tested all the buttons just now and they all work too, buttons, limits (SoC limit slider), they all work correctly on the Delta 3 Classic.

shuette42:
Thanks for testing the controls, much appreciated. That confirms full support for the Delta 3 Classic in v1.15.0-beta.10.

PR #134, opened 2026-07-23
Feature/bkxx stream support

davidpicon7:
## Related Issue

Closes # https://github.com/shuette42/ecoflow-energy-ha/issues/71

## Summary

This PR builds on the Stream AC Pro support introduced in https://github.com/shuette42/ecoflow-energy-ha/commit/8b688c635ba9580b449a2fc6554306a7916a6d71 and adds support for the remaining Stream devices that were still missing: Stream AC, Ultra, Ultra X, and Max.

It reuses the existing entities. I have two Ultra units and one Ultra X at home, and all of the previously implemented entities work as expected.

## Changes

- Add new devices at const.py and modify the way to obtain their name    
    BK-series Stream devices:
    BK11: Stream Ultra
    BK31: Stream AC Pro
    BK41: Stream Max
    BK51: Stream AC
    BK61: Stream Ultra X

## Checklist

- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated

## Screenshots
- Tested on HA
image
- Tests
-  
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.15.1] - 2026-07-23
+
+### Added
+- Added serial-prefix and display-name support for the additional Stream family models Stream Ultra (BK11), Stream Max (BK41), Stream AC (BK51), and Stream Ultra X (BK61) so they are recognized correctly in Enhanced Mode and shown with their proper model names in Home Assistant.
+
 ## [1.15.0] - 2026-06-14
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,6 @@
 
 All notable changes to this project will be documented in this file.
 
-## [1.15.1] - 2026-07-23
-
-### Added
-- Added serial-prefix and display-name support for the additional Stream family models Stream Ultra (BK11), Stream Max (BK41), Stream AC (BK51), and Stream Ultra X (BK61) so they are recognized correctly in Enhanced Mode and shown with their proper model names in Home Assistant.
-
 ## [1.15.0] - 2026-06-14
 
 ### Added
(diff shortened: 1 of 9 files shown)


shuette42:
Thanks, merged. Having someone with actual Ultra and Ultra X hardware in one install is the reason this could go in at all, testing against real devices is exactly what the STREAM family was missing.

**What works now:** serial-prefix recognition and display names for BK11 (Stream Ultra), BK31 (Stream AC Pro), BK41 (Stream Max), BK51 (Stream AC) and BK61 (Stream Ultra X).

**Two adjustments I made on top:**

**Naming scope.** The prefix-derived friendly name plus serial tail applied to any device where the API returns an empty product name. That also catches PowerOcean, Delta 3 and Smart Plug, so those devices would have changed their labels for existing users. Not something visible from a STREAM-only setup. I restricted the naming to the BK series so existing devices keep their current labels.

**Version convention.** The repo is mid-cycle on 1.15.0, currently at beta.12, so 1.15.0 is not released yet. Repo convention: during an open beta cycle the manifest stays at the target version and the CHANGELOG keeps one block per target version. So the manifest is back at 1.15.0 and the entry folded into the existing 1.15.0 block. For a future PR that just means leaving the manifest version alone unless the PR opens a new cycle, and adding the entry under whatever version block is already at the top.

Minor cleanup on top: removed a duplicated display-name map, a no-op branch in the serial-tail logic, and some unused imports.

**Verification:** full test suite passes, and I checked it in a local Home Assistant instance. The integration loads clean and existing PowerOcean entities are not renamed.

**Open gap:** BK11 and BK61 are hardware-verified through your setup. BK41 and BK51 are not verified by anyone yet. Recognition and naming should be correct, but the telemetry field layout currently assumes it matches BK31. Tracked separately in #136.

This is on main and ships in v1.15.0-beta.13.

Two things if you have time: once beta.13 is out, it would help if you re-check your Ultra and Ultra X setup, mainly that the device names still look right after the scoping change and nothing else got renamed. And if you want to keep contributing on the STREAM side, the door is open, you are the only person so far running two of these models in one install.

davidpicon7:
Thanks for the incredibly fast review and for merging this so quickly!

Also, thank you for maintaining this HACS integration. In my experience it works much better than the official one, so I really appreciate the work you've put into it.

I've just tested it by installing v1.15.0-beta.14, and I can confirm everything is working correctly on my setup. The device names look right after the scoping changes, and I haven't noticed anything else being renamed.

I'll probably submit a few more improvements, mainly focused on the Stream Ultra and Stream Ultra X, since those are the models I own and can test with real hardware. In particular:

Add support for exposing the energy generated by each individual PV panel (the official integration already provides this).
Make the AC1 and AC2 controls actual switches instead of read-only entities, since the official integration also allows controlling them.

Hopefully these improvements will be useful for other STREAM users as well.

shuette42:
Thanks for verifying on real hardware. Your beta.14 check on the two Stream Ultra and the Stream Ultra X is exactly what was needed: device names correct after the scoping change, and nothing else renamed.

PRs for both improvements are welcome. I've opened two tracking issues so the work has a home:

- Per-panel PV energy: #139
- AC1/AC2 switch control: #140

Attach your PRs to those and it'll be easy to follow.

A few notes for when you open them:

1. Stay in the 1.15.0 beta cycle, don't bump the manifest (same convention from earlier in this thread).
2. Since you run Ultra and Ultra X, verify the changes on that hardware. BK41/BK51 stay out of scope until someone can verify their telemetry (tracked in #136).
3. AC1/AC2 is a write path, so the SET command per output needs confirming before it ships. That one is the trickier of the two.

Happy to review both once you have them up.

Issue #88, opened 2026-05-20
Power Ocean Plus (30KW)


TimBreaksStuff:
### Device Type

PowerOcean

### Problem or Use Case

Hi, I used this plugin a lot and really love/ enjoy it! However, over the weekend I change my PowerOcean (10KW) to the Power Ocean PLUS (30KW) and does not pick up any new data. 

It would be super helpfull if this plugin would also support the PowerOcean PLUS. 

If any information from my end is needed, LMK

Thanks again,
Tim H.

### Proposed Solution

Add option for Power Ocean PLUS

CapeH02:
I'm experiencing the same issue: I was recommended the integration, thank you for the initiative. 
However, after a fresh installation and successful login with device selection, no entities are displayed (PowerOcean Plus, 15kWh battery, enhanced mode). Perhaps an access setting needs to be enabled by the electrician?
Thanks, Cape

TimBreaksStuff:
I have the "installer" app and there is no direct option to enable this. I can enable API, but even then, no data/ values are coming through.

Xygen:
@TimBreaksStuff Oh! How did you gained access? 🤩

TimBreaksStuff:
Its super easy, you go on their website and registrate as a "installer" :)

Xygen:
That was easy indeed... Wow!
But it looks like my installer needs to remove it from their account first :/

henhappm:
Hi @shuette42, thanks a lot for this integration — really impressive work. Adding a +1 here with another PowerOcean Plus variant, in case it helps with the mapping.

I'm hitting the same "no parser available" skip (same symptom as #89) on a PowerOcean Plus, inverter P3-S1 (25 kW, 3-phase hybrid). Setup is otherwise fine: 2FA off, Enhanced mode, the inverter is discovered and selectable, login and discovery succeed — but no entities are created.

Debug log (custom_components.ecoflow_energy at DEBUG):

    DEBUG [custom_components.ecoflow_energy] Skipping device R374... - no parser available
    DEBUG [custom_components.ecoflow_energy] EcoFlow Energy: 1 device(s) configured (Enhanced: 1, Standard: 0)

Details:
- Model: EcoFlow PowerOcean Plus, P3-S1 (25 kW, 3-phase)
- Serial prefix: R374ZD1AZH81…
- Mode: Enhanced
- EcoFlow Energy: 1.14.0 (HACS, latest)
[...]

PR #157, opened 2026-07-26
fix: decode every PowerOcean protobuf header

henhappm:
## Status

Ready for maintainer review. This is the hardware-validated R374 parser fix requested in #88.

## Root cause

The runtime decoder selected the command tuple and `pdata` from only the first header of an MQTT envelope. PowerOcean Plus Get-All replies contain multiple independent headers, so everything after header 0 was silently lost. The same device also continues to send encrypted incremental pushes.

## Approach

The change is additive to the shared PowerOcean path:

- iterate every recognized header in bundled replies and incremental pushes;
- decrypt each encrypted `pdata` independently with that header's own `seq & 0xff` key;
- merge decoded mappings from one MQTT envelope;
- ignore unsupported and empty companion headers;
- keep single-header PowerOcean devices on the same code path, with no separate Plus branch;
- retain the existing first-result runtime API and Get-All wrapper for compatibility.

On the tested R374, Get-All arrives as a repeatable 19-header bundle. Incremental traffic includes `(96,33)+(96,39)` pairs, isolated `(96,1)` heartbeats, and additional `(96,8)`, `(96,16)` and `(96,17)` updates. `(96,39)` had zero-length `pdata` both at night and during active production; live flow values were carried by `(96,33)`.

## Additional parser fixes

- flatten all EMS MPPT heartbeat containers, exposing up to four physical PV inputs;
- zero-fill omitted proto3 power/voltage/current scalars inside heartbeat snapshot entries that actually exist;
- expose reactive and apparent phase power already present in `JTS1PhaseInfo`;
- preserve support for non-empty `(96,39)` variants.

## Hardware validation

Tested on a real EcoFlow PowerOcean Plus R374 without a battery in Enhanced Mode and compared live against a second working PowerOcean Plus integration on the same Home Assistant instance.

During approximately 6.7 kW production:

| Measurement | This parser | Reference |
|---|---:|---:|
| Total PV | 6730 W | 6730.0 W |
| House power | 450 W | 450.0 W |
| Grid power | -6280 W | -6280.0 W |
| MPPT 1 | 3529 W | 3528.7 W |
| MPPT 2 | 1772 W | 1771.6 W |
| MPPT 3 | 1436 W | 1436.4 W |

Night validation also matched at 0 W total PV and 0/0/0 W across the installed MPPT inputs, with no stale low-light values.

After removing the temporary capture hook and restarting Home Assistant again, the integration was rechecked as loaded/healthy, WSS connected and receiving, 47 data keys, and 0 reconnects.

## Real masked fixture

The requested full real-device fixture is included:

- `tests/fixtures/powerocean/r374_get_all_masked.bin`
- 2,483 bytes
- 19 independent headers
- SHA-256 `9663864d26273144f8564f284a376ae26ba9dd76056ccf68db63fe85a85dcaf2`
- four 16-character device/module identifiers replaced with equal-length placeholders
- no MQTT topic or credentials stored
- post-mask checks found no EcoFlow-style serial, long ASCII account identifier, or email pattern

The companion JSON contains only non-sensitive command metadata and privacy-check results.

## Tests

Regression coverage verifies:

- an unknown first header does not prevent later known headers from decoding;
- different encrypted headers use their own sequence numbers;
- bundled energy-stream and EMS-heartbeat data are both retained;
- empty known companion headers are ignored safely;
- the legacy single-header decoder API remains compatible;
- four MPPT inputs are flattened across multiple heartbeat containers;
- nested proto3 zero values clear previous MPPT/phase readings;
- reactive/apparent phase power is preserved;
- coordinator-level mappings from a bundle are merged into one update;
- the real masked fixture remains 19 headers and reaches EMS heartbeat, EMS change and energy-stream parsers.

## CI status

- [x] Fork `Tests` workflow completed successfully on the exact real-fixture commit
- [x] Hassfest completed successfully
- [ ] HACS validation: the fork-only action fails in fork repository context; upstream validation remains authoritative
- [ ] Upstream workflows await maintainer approval for this first-time fork contribution
- [ ] CHANGELOG entry after maintainer confirms the release/beta target

## Compatibility and limitations

- HJ31/HJ32 hardware was not available locally, but those variants use single-header frames and continue through the same shared path.
- No physical battery was installed, so positive battery-pack behavior could not be hardware-tested.
- R371 and HJ3C remain unconfirmed.

Refs #88.


henhappm:
The requested full masked R374 Get-All capture is now in the branch as a real regression fixture:

- raw binary frame: `tests/fixtures/powerocean/r374_get_all_masked.bin`
- non-sensitive metadata: `tests/fixtures/powerocean/r374_get_all_masked.json`
- 2,483 bytes / 19 headers
- SHA-256: `9663864d26273144f8564f284a376ae26ba9dd76056ccf68db63fe85a85dcaf2`
- four 16-character device/module identifiers replaced with equal-length placeholders
- no MQTT topic or credentials stored
- post-mask scan found no EcoFlow-style serial, long ASCII account identifier, or email pattern

The fixture test confirms that the outer decoder still sees all 19 headers and that the shared runtime path reaches EMS heartbeat, EMS change, and energy-stream decoding. After removing the temporary capture hook and restarting Home Assistant again, the real R374 was healthy with 47 data keys and 0 reconnects.

The fork-only CI rerun for this exact head is complete: the full `Tests` workflow and Hassfest passed. The HACS action still fails only in fork repository context, so the upstream repository run remains the authoritative HACS check. The PR is now ready for maintainer review, and the upstream workflows are awaiting maintainer approval for this first-time fork contribution.

shuette42:
Merged. Thanks for this, and specifically for the two things that made it reviewable: the cross-check against a second working PowerOcean Plus integration on the same instance, and the masked raw capture. The capture is the reason a proper before/after comparison was possible at all. With it I could run all seven PowerOcean frame kinds through the old and the new path and compare the output directly instead of reasoning about the diff.

Your core design went in as it was: iterate every header in a bundled reply instead of reading only the first. That is the right structure for the observed device behaviour.

It also fixes a second problem. In the old path the command function and the command id could come from two different headers of the same bundle and end up mismatched, so a frame could be routed to the wrong parser. I had run into that here and never wrote it up anywhere public. Your structural change rules it out because both values are now read off the same header. That one is yours, it falls out of the design.

The regression result: seven PowerOcean frame kinds traced old against new, five identical, two producing strictly more data. Then confirmed on real HJ31 hardware in a running Home Assistant instance, same entities, same values, nothing in the log.

The fixture stays in the repo.

Two things needed fixing before merge, and neither is a flaw in your change:

A shared decoder path that the new PowerOcean flow no longer uses, but that Delta 3 and Smart Plug still depend on, had lost a decode step along the way and returned no values at all for certain multi-header frames. Restored in the same pass.

The CHANGELOG entry, which you correctly left open since the release target is my call. Added.

A couple of edge cases in the merging of bundles got tightened at the same time: an all-zero voltage row could flip the grid status reading, and a single unusable header could discard an otherwise good bundle. Both are handled now.

All of that is my side of the split we agreed on, so there is nothing for you to touch.

It is out now as v1.15.0-beta.22, so you can run it on your own R374.

One question, and it is genuine curiosity rather than a task: have you ever seen `enc_type` set to 1 on the Plus, or was it absent in every frame you captured? Every header in the capture you shipped leaves it out, so the only coverage for that path here is synthetic.

henhappm:
Thanks again for the merge and for the very thorough review.

To answer your `enc_type` question: it was absent in every header of the 19-header bundled Get-All capture. On the same R374 I now see `enc_type=1` explicitly on live incremental property frames, including `(96,33)+(96,39)`, `(96,8)`, `(96,1)` and `(241,5)`. So the Plus uses both forms: omitted in the bundled Get-All reply, but present as `1` in incremental pushes.

While validating beta.23, I found one small follow-up that is on me: the local prototype I tested already had the MPPT 3 entity definitions and translations, with MPPT 4 defined but disabled by default, but I missed including the entity-layer part in PR #157.

The decoder itself is working correctly: diagnostics contain `mppt_pv3_power_w`, `mppt_pv3_voltage_v` and `mppt_pv3_current_a`, and the pre-merge prototype recorded live MPPT 3 values for many hours. In beta.23, however, `POWEROCEAN_SENSORS` and the translations still stop at MPPT 2, so the existing MPPT 3 registry entries become stale/restored and unavailable after restart.

I opened #160 as a small follow-up. It restores the MPPT 3 entities, keeps MPPT 4 disabled by default, adds the English and German translations, updates the entity documentation/counts, and adds regression coverage tying parser-supported MPPT inputs to their entity definitions.

Sorry I missed that part in the original PR, and thanks again for the careful review.

henhappm:
Thanks again for the merge and for the very thorough review.

To answer your `enc_type` question: it was absent in every header of the 19-header bundled Get-All capture. On the same R374 I now see `enc_type=1` explicitly on live incremental property frames, including `(96,33)+(96,39)`, `(96,8)`, `(96,1)` and `(241,5)`. So the Plus uses both forms: omitted in the bundled Get-All reply, but present as `1` in incremental pushes.

While validating beta.23, I found one small follow-up that is on me: the local prototype I tested already had the MPPT 3 entity definitions and translations, with MPPT 4 defined but disabled by default, but I missed including the entity-layer part in PR #157.

The decoder itself is working correctly: diagnostics contain `mppt_pv3_power_w`, `mppt_pv3_voltage_v` and `mppt_pv3_current_a`, and the pre-merge prototype recorded live MPPT 3 values for many hours. In beta.23, however, `POWEROCEAN_SENSORS` and the translations still stop at MPPT 2, so the existing MPPT 3 registry entries become stale/restored and unavailable after restart.

I opened #160 as a small follow-up. It restores the MPPT 3 entities, keeps MPPT 4 disabled by default, adds the English and German translations, updates the documentation, and adds regression coverage tying parser-supported MPPT inputs to their entity definitions.

Sorry I missed that part in the original PR, and thanks again for the careful review.

PR #169, opened 2026-07-31
feat: add PowerGlow heating rod power sensor

Xygen:
## Summary

- Add the PowerGlow quota field `ems_heating_rod.heatingPower` to the existing PowerOcean HTTP quota parser.
- Expose it as the read-only `Heating Rod Power` PowerOcean sensor, disabled by default.
- Add anonymized fixture coverage using a dummy PowerOcean serial, translations, and a CHANGELOG entry.

## Scope

This deliberately contains no new device type, serial-prefix detection, accessory polling, MQTT changes, controls, or power-to-energy integration.

## Validation

- Observed through the PowerOcean quota path while idle: `ems_heating_rod.heatingPower = 0`.
- Parser checks cover a valid fixture value, an absent field (which emits no placeholder zero), and a non-numeric value.
- Python compilation, JSON parsing, translation/fixture checks, and `git diff --check` pass locally.
- The full pytest suite could not be run locally because the available Python runtime is older than the project's required Python 3.13 test environment.

## Before merge

An active-heating sample, paired with the simultaneous EcoFlow app power reading, will be added once the buffer tank temperature permits a representative heating cycle. This confirms the field scaling before the sensor is merged.


shuette42:
Reading the accessory out of the PowerOcean quota instead of building a separate device type is exactly the shape I want for this class of hardware, and you got there without being told.

On the scaling: I checked independently and plain watts holds up. The app renders this reading without any arithmetic on it and only switches the displayed unit to kW once the value passes 1000, and the neighbouring fields in the same payload (temperature, tank volume) are unscaled as well. That only confirms what the app does with the value, not what the cloud quota returns for it, so your non-zero comparison is still the piece that closes the gap. I'm waiting on it the same as you are.

Two things I added on top while integrating:

**Test with a non-zero value.** Your three cases all run against the observed idle zero, which passes through unchanged no matter what scaling the parser applies. A non-zero case pins the unit contract, so a divisor slipping in later fails loudly instead of silently.

**Version and docs.** Resolved the CHANGELOG date conflict, moved the entry into a new 1.16.0 block, bumped the manifest, and corrected the PowerOcean sensor count from 208 to 209 in the four places it appears. You're credited by name in the CHANGELOG entry.

Two asks, different timelines.

The first needs no waiting: download diagnostics while the system is idle and post everything under the `ems_heating_rod.*` key. You already see `heatingPower = 0` at idle, so whatever else sits under that key should be there right now. That tells us what else the accessory reports (temperature, run status, target values) and whether a second sensor is realistic.

The second is the one you already proposed under "Before merge": the non-zero reading with the simultaneous app value. No deadline on that, the buffer tank decides when it happens.

If the dump shows a usable temperature reading I'd want to look at it, but I'm not committing to anything before seeing what's actually in there.

shuette42:
Correction on my first ask: the diagnostics dump I asked you for was not possible. The raw quota section only existed for two other device families, and for PowerOcean that data was never collected in the first place. If you went looking for `ems_heating_rod.*` and found nothing, that was on my side, not yours.

That is fixed now in **v1.16.0-beta.2**: https://github.com/shuette42/ecoflow-energy-ha/releases/tag/v1.16.0-beta.2

**What changed:**

1. A diagnostics download now contains the raw quota keys and values for PowerOcean, with serial numbers masked. The original ask is actually doable now.
2. Three more readings next to the power value: water temperature, target power, target temperature. All four are read-only and disabled by default, so you need to enable them in the entity settings of the PowerOcean device.

On expectations for the three new ones: I don't know whether they will get values. The field names for the heating rod are not documented and differ between systems, so the water temperature is read under three possible names. If the sensors stay empty, that is not a defect, it is exactly the information that is missing right now. Please report empty sensors as a finding, not as a bug.

Two asks, in this order:

1. **Update to beta.2 and attach a diagnostics download.** Beta channel has to be enabled in HACS. This is the single most useful thing you can contribute right now, because the dump shows which fields your heating rod actually delivers, and that decides what can be built next.
2. **Enable the four sensors and tell me which ones show values and which stay empty.** For the ones with values: do they match the EcoFlow app? Temperatures should come out as whole degrees.

The reading under load is still welcome whenever the buffer tank gets there.

PR #204, opened 2026-08-06
feat: add EcoFlow STREAM AC 5000 (ES22) read support

slaapyhoofd:
Adds read support for the EcoFlow STREAM AC 5000 (serial prefix `ES22`): its own device
type, parser and entity set, 52 sensors and two binary sensors.

Read side only, as agreed on #177. The controls are written and hardware-tested but stay
out of this PR so they can carry their own evidence into their own review.

## Why it needs its own parser

An ES22 shares the Stream product name and nothing at the protocol level. It sends no
`254/21` frame; its telemetry is `254/39` and `254/40`, and it describes power as a matrix
of flows between grid, battery, house and solar, nested two to five levels deep.
`_STREAM_FIELD_MAP` is keyed on top-level field numbers and cannot address `f11.1` or
`f40.1.9.1`. Routing it to `DEVICE_TYPE_STREAM` would also have put roughly fifteen
ES22-only keys into `STREAM_SENSORS`, and since `_SN_PREFIX_EXCLUDED_KEYS` is subtractive,
those entities would have been created on all six existing BK prefixes permanently.

`32/50` is common to both and its key names are kept identical, with one exception: field
25 is the BMS pack SoC on an ES22 and runs about two points above the system SoC the app
shows, so it gets its own key rather than sharing `soc_precise_pct`.

## Evidence

A 1239-frame capture of a live unit in Enhanced mode alongside the EcoFlow app, plus the
diagnostics attached to #177 from a second unit with different hardware and a different
meter. Cross-checks that hold across the capture:

- house consumption equals the sum of the flows into the house, within 2 W on every push
  frame
- the grid node equals the sum of the grid flows, worst case 1 W
- the grid reading agrees with an independent Tibber Pulse on the same connection, median
  0 W

Scaling proved against the app at the same moment: `f11.1` = 4042 halves to 2021 W against
2021 W in the app, `f11.9` is the one field in that block already in watts, and `f10.1`
read 2400 the instant the app's output limit was set to 2400 W.

## Two decisions worth a look

**Zero-filling the flow edges.** An omitted scalar inside a message that was sent is zero
by definition, and none of 1024 observed `f12` subfields was ever an explicit zero. The
fill applies only when the container is present. An absent `f12`, 116 of 355 pushes, means
unchanged and clears nothing. No filled key is `total_increasing`.

**Battery power comes from the flow edges, not from the device's own `f50.1.4`.** That
field is accurate while the unit is switching, but the whole `f50` block stops being sent
the moment the flow stops, so it latches: measured at 444 W of "charging" for over five
minutes after an external AC meter had gone to zero and the pack had begun discharging.
`(f12.7 + f12.9) - (f12.4 + f12.5)` tracks that meter within about a watt charging,
discharging and at rest. Detail in #177.

## Verification

1828 tests pass, hassfest clean. Read entities confirmed on live hardware, including the
accessory gating: the solar entities appeared on their own once the device reported them,
while the per-phase meter entities stayed absent on a unit with a single-total meter.

Ref #177


shuette42:
ES22 read support is on main, and it is out as `v1.17.0-beta.1`. It went in as #208, which carries your commit unchanged as its first commit.

The reason it is a second pull request and not this one is a GitHub mechanic rather than a decision: your fork sits under an organization, and organization-owned forks do not grant maintainer edits. There was no way to push follow-up commits onto your branch, so the branch was rebuilt with your commit at the base of it and the corrections on top. It was not squashed into anything. It stands on its own in the history under your name.

What you brought here is worth naming precisely, because "thanks for the PR" does not cover it. You captured a live unit alongside the app and cross-checked the grid against an independent meter on the same connection, which is the only reason any of the numbers could be trusted rather than believed. The flow matrix is your model, and it came with in-frame balances, every term of each equation from the same message, so no carried-over value could flatter the result. And separately, you caught the latching `f50.1.4` battery field yourself, in your own analysis on the issue thread, before this pull request existed. You drove the unit idle to charging to idle with a meter on the AC side and watched it hold 444 W five minutes after the flow had stopped. That one was yours. Nobody found it for you.

Two things changed on top of your work, and both are decisions about how energy counters behave across this integration rather than anything wrong on the read side, which was accurate:

- The lifetime solar counter is gone. The live solar reading stays. The device works out a solar figure for itself and reports it even with no PV wired to it, and a counter that only ever counts up cannot be corrected once it has counted.
- The SoC limits are read from one frame instead of two. Both carried the same values, so nothing was flapping, but two sources for one entity is the shape that made the Stream LED readback flap.

A third change was made and then taken back, and you should know about it because the history shows it. The solar-to-battery edge was pulled out of the battery derivation on the grounds that it appears in none of the 31 frames shipped as fixtures. It appears in none of them because those frames come from your unit, which has no PV attached, and you said exactly that when you sent them. Your own table on the issue thread has the field with values, and the 12:19 frame settles it: 46 against a battery taking 47 W out of 2.86 kW of solar, where the opposite assignment would have needed 2860. It is back in, with those readings written next to the mapping instead of a guess. Absence from a capture only means something when the capture could have contained the thing, and that is on me rather than on your work.

Two smaller corrections while in there: the grid import and export icons pointed the opposite way to the PowerOcean definitions of the same four keys, and the entity page described only one of the two task-deletion cases.

The control side you held back is yours to bring as its own pull request whenever it suits you, on the same terms as this one. No date needed from you, and none promised from me on when I get to it.

PR #208, opened 2026-08-06
feat: add EcoFlow STREAM AC 5000 (ES22) read support

shuette42:
Read support for the EcoFlow STREAM AC 5000 (serial prefix `ES22`): its own device type, parser and entity set, 51 sensors and two binary sensors.

The work is slaapyhoofd's, contributed as #204 and carried here unchanged as the first commit, with the review corrections on top. It could not go onto his own branch because the fork sits under an organization and does not grant maintainer edits.

## What the review changed

**The lifetime solar counter is gone, the live reading stays.** The device works out a solar figure of its own from the house flows and reports it on units with no PV wired to it. Frames 25 and 26 of the shipped capture show exactly that, on a unit with no panels on the EcoFlow. As an instantaneous reading it is the device's own number and belongs in the integration. As a `total_increasing` counter it would put production that does not exist into the Energy Dashboard, and a counter of that kind only ever goes up, so nothing corrects it afterwards. The entity page said the reading appears "only on a unit with PV wired to the EcoFlow" in its table and the opposite in its notes; README and changelog repeated the table. All three now say what the device does.

**The battery term keeps the solar-to-battery edge, after a correction to the correction.** The first review took field 9 out because it appears in none of the 31 captured frames. It appears in none of them because the unit they came from has no PV wired to the EcoFlow, which its owner stated when he sent them. On #177 the field is confirmed three times on a unit that has PV, the clearest being a frame where the battery took 47 W out of 2.86 kW of solar and field 9 read 46. Left out, a PV owner would see zero battery power while the pack fills.

**The SoC limits come from one frame instead of two.** Both carried identical values, so nothing flapped, but two sources for one entity is the shape that made the Stream LED readback flap.

Also corrected: a comment that claimed the zero-fill contains a wrong field guess and does not, three mappings resting on a live session rather than on the captures, and one key that broke the percent naming convention.

## Tests

2154 pass, up from 2145. Three that were missing:

- the six BK prefixes gain and lose no entities from this device type, which is the failure the separate device type exists to prevent
- no serial prefix may be added for a family the not-this-family guard rejects, since the prefix map now answers before that guard runs
- the two frames where a PV-less unit reports solar still parse as they do

## Verification

Docker gate green at `153a9c7`: integration loads, no errors, no new warnings. The only warning is a device on the test account that has no parser yet and is unrelated to this change.

No ES22 hardware here. The readings themselves rest on slaapyhoofd's capture of a live unit alongside the app, cross-checked against an independent meter on the same connection, and on the diagnostics from a second unit with different hardware and a different meter attached to #177.

Ref #177


slaapyhoofd:
Thanks for carrying this and for reviewing it that closely. The solar counter was the
right call and I should have caught it myself: my own entity note said the device
infers solar and reports it with no PV wired, and I shipped a `total_increasing`
counter integrating it anyway.

I took the three "this rests on a live session" caveats as a prompt to put them on
hardware. Four are now confirmed. Two other things need correcting.

## The SoC limits: `32/2` is the frequent source, `254/39 f29` the rare one

Changed Max Charge SoC from 100 to 95 in the app while polling diagnostics every 6 s:

```
16:12:42   32/2    f1.7    100 -> 95    size 114
16:17:09   254/39  f29.1   100 -> 95    size 192
```

Four minutes 27 seconds apart. Over the same 13-minute window, with identical buffer
retention, `32/2` carried the limits **83 times** and `254/39 f29` carried them **4
times**.

The principle is right, one source per entity. But `f29` only travels in the
~1400-byte full-state frame and in occasional small config pushes, while `32/2`
carries the limits in every frame it sends. Keeping `254/39` as the survivor means a
limit changed in the app takes minutes to reach Home Assistant instead of seconds.

Worth adding: the two never disagreed. Four distinct value pairs across the captures
(90/15, 90/20, 100/10, 95/10), both sources matching every time. The flap you were
guarding against is not present on this device.

## `10.6` is not the input limit, `10.2` is

This one is my bug rather than a review one. Set the app's Max grid input power to
1200, then back to 2500, forcing a fresh get-all each time:

| app input limit | `f10.1` | `f10.2` | `f10.5` | `f10.6` |
|---|---|---|---|---|
| unknown, 2026-08-03, ceiling 600 | 600 | 2500 | 600 | 600 |
| 2500 | 2500 | 2500 | 800 | 2500 |
| **1200** | 2500 | **1200** | absent | absent |
| **2500** | 2500 | **2500** | 800 | 2500 |

Only `.2` moves with it, in both directions. `.1` follows the grid-tied output limit,
confirmed separately by moving it to 1500 and back. `.6` has moved once ever, when the
account ceiling went from 600 to 2500, and did not move through either user change, so
it behaves like a ceiling.

So `max_grid_input_power_w` should read `10.2`. As shipped it reports the output
ceiling under the input limit's name, which looked correct here only because both sat
at 2500. That also inverts the note added in the review: `f10.2` is the user setting,
not the hardware ceiling.

`10.5` stays unexplained. It went from 600 to 800 when the ceiling was raised, and
matches no setting visible anywhere in the app. It stays unmapped.

## The four that are now on hardware

| field | values seen | how |
|---|---|---|
| `19.1` | 0 and **1** | backup socket toggled twice in the app |
| `30.1` | 0 and **1** | backup reserve switched on, then off |
| `30.2` | 40, **35**, **65** | reserve level moved twice |
| `10.1` | 2500, **1500** | grid-tied output limit changed and restored |

Each moved its entity within seconds, so the mappings and the paired-write behaviour
both hold.

## Why the fixtures could not show any of this

The frame buffer keys on topic plus command and keeps three per bucket, while pushes
arrive every two seconds. A single diagnostics pull therefore yields one frame per
message type, carries no arrival-rate information at all, and drops most config
pushes. Everything above needed repeated polling or Home Assistant's own state
history.

That is worth knowing beyond this device: the capture artifact is good for building a
parser and cannot answer "how often". Both misses above are downstream of it, not of
the reading.

I can ship a replacement fixture spanning the config changes above, masked the same
way, so the arrival rates live in the repo rather than in this comment.

PSCourtney:
The Stream 5000 is ES21 btw. 
image

PR #210, opened 2026-08-06
fix: correct two ES22 config readbacks confirmed on hardware

slaapyhoofd:
Two read-side corrections for the STREAM AC 5000, both from a session that moved each
setting in the EcoFlow app and watched what the device reported. Both affect
`v1.17.0-beta.1`. Details and measurements are in #208.

Separate commits, so either can be taken without the other.

## `f10.6` is not the grid input limit, `f10.2` is

`max_grid_input_power_w` was reading a field that does not follow that setting. The
app's "Max grid input power" was set to 1200 W and back to 2500 W, with a fresh
get-all at each step:

| app input limit | `f10.1` | `f10.2` | `f10.5` | `f10.6` |
|---|---|---|---|---|
| 2500 | 2500 | 2500 | 800 | 2500 |
| **1200** | 2500 | **1200** | absent | absent |
| **2500** | 2500 | **2500** | 800 | 2500 |

Only `.2` moves with it, in both directions, and nothing in 452 captured frames ever
held 1200. `.1` follows the grid-tied output limit, confirmed in the same session at
1500 W and back. `.6` has moved once in four days of captures, when the account
ceiling went from 600 to 2500, and it did not follow either user setting.

Both sat at 2500 on the unit the mapping was built from, which is the normal state
until someone lowers the charging limit, so the wrong field read correctly right up to
the point where it mattered.

## The SoC limits come from `32/2`, not `254/39 f29`

Keeping one source is right. `f29` is the wrong one to keep. Max Charge SoC changed
from 100 to 95 in the app, polling every 6 s:

```
16:12:42   32/2    f1.7    100 -> 95    size 114
16:17:09   254/39  f29.1   100 -> 95    size 192
```

Four minutes 27 seconds apart. Over the same thirteen-minute window, identical buffer
retention: `32/2` carried the limits 83 times, `254/39 f29` carried them 4. `f29`
rides only in the full-state frame and in the occasional small config push after a
change.

The two never disagreed, across 90/15, 90/20, 100/10 and 95/10, so the flap this
guards against has not been seen on this device. The Delta 3 divergence is deliberate
and recorded at the mapping.

## Also

Two mapping comments were retired because the same session disproved them: `19.1` said
the switched-on branch had never been seen on hardware, and `30.2` said the reserve was
only ever observed at a constant 40. The backup socket was toggled on and off twice,
and the reserve moved 40 to 35 to 65 to 40 with the enable flag switched around it.

## Tests

2163 pass, up from 2162 on main. Two existing tests swap sides, one is new: `f10.5`
and `f10.6` together must not produce an input limit.

No new fixture in this PR. The frame buffer keeps three frames per topic-and-command
bucket while pushes arrive every two seconds, so a diagnostics pull cannot show
arrival rates and drops most config pushes, and the stored hex is truncated at 512
bytes. A fixture worth shipping needs the untruncated capture path and its own masking
pass. Happy to do that as a follow-up if you want the rates in the repo.

Ref #177


shuette42:
Merged by rebase, so both commits stand on main under your name. Released a few minutes ago as `v1.17.0-beta.2`, pre-release.

Thanks for the method more than the patch. An hour after beta.1 went out you had moved every setting in the app and watched what came back, and that is the only thing that could have caught either of these.

**The SoC limit source was our call, not yours.** The review found two sources for one entity and kept `254/39 f29`. The principle was right and the survivor was wrong. 83 arrivals against 4 over thirteen minutes, and a change made in the app reaching the kept source four minutes twenty-seven seconds late, settles it. `32/2` is the source.

The input limit was yours from beta.1, and it is the kind that reads correctly until it matters: both fields sit at the same value until someone lowers the charging limit, so nothing on your unit could have shown it until you lowered it.

**On the four caveats:** putting `19.1`, `30.1`, `30.2` and `10.1` on hardware instead of accepting three "this rests on a live session" notes was the right response. They are confirmed now rather than flagged.

**Follow-up pass:** #213 is open and touches your files, so you are not surprised by it. Comments and tests only, no behaviour, which is why the release did not wait for it. It cannot merge yet because GitHub Actions is in a major outage and the required checks will not start. What is in it, in case you want to object to any of it: the comment calling both `f10.5` and `f10.6` account ceilings is narrowed to `f10.6`, since `f10.5` is unexplained rather than a ceiling; the sensor comment still described the input limit as account-level; a replay test still allowed a `32/2` frame to parse to nothing, which is a hole now that the limits ride there; and the third copy of the SoC limits at `254/39 f33.7`/`f33.8` is marked as deliberately unmapped.

**One thing a follow-up review turned up that I would value your view on.** The solar paths `11.9` and `12.8` are held out of the zero-fill so the accessory gating holds. Zero-fill is also the only thing that turns an omitted scalar into a real zero, and nothing deletes a key that stops arriving, so on a unit with PV the solar reading holds its last daylight value all night. The push capture has the `f11` container arriving fourteen times with `11.9` omitted, so it is the ordinary case rather than an edge one. On a unit without PV it lands differently: the device infers a solar figure in two of the thirty captured frames, and one such frame is enough to create the entity permanently, enabled, where it then sits frozen.

Curing it means publishing zeros, which creates the solar entities on every unit, which is what the gating exists to prevent. So it is a design decision rather than a patch, and I am working through the options now: gate the accessory on a value ever being non-zero rather than on the key appearing, do the filling a level up where the device state knows what has been seen before, or make the accessory decision explicit per key. If you have a preference, or a reason one of those is worse than it looks, now is the moment.

**Yes to the fixture.** A replacement spanning the config changes, masked the same way, is worth having. The premise of the second correction currently exists only as prose in a pull request. If the arrival rates live in the repo, the next person can check the ranking instead of trusting it.

Your closing observation is the most useful thing in this exchange. A capture artifact is good for building a parser and cannot answer "how often", so it cannot rank two sources by freshness and cannot separate two fields that happen to hold the same value. Both misses are downstream of that, not of anyone's reading. I am taking that as a general lesson for the repo rather than an ES22 note: any mapping that survives a comparison needs a rate behind it, not just a match.

PR #212, opened 2026-08-06
fix: correct two ES22 config readbacks confirmed on hardware

shuette42:
Two read-side corrections for the STREAM AC 5000, both measured on live hardware by slaapyhoofd within an hour of `v1.17.0-beta.1`. His two commits are carried here unchanged, with a review pass on top. They could not go onto his own branch because the fork sits under an organization and grants no maintainer edits, the same as #204.

Contributed as #210.

## `max_grid_input_power_w` read a field that does not follow that setting

He moved the app's "Max grid input power" to 1200 W and back to 2500 W with a fresh get-all at each step. Only `f10.2` follows, in both directions, and nothing in 452 captured frames ever held 1200. `f10.6` has moved once across four days of captures, when the account ceiling went from 600 to 2500, and through neither user change.

Both fields sat at 2500 on the unit the mapping was built from, so the wrong field read correctly right up to the point where it mattered.

Corroborated from an artifact already in this repo: both shipped fixtures decode to `10.1=600, 10.2=2500, 10.5=600, 10.6=600`, which is his 2026-08-03 row field for field. That same decode disproves the note the previous review added: if `f10.2` were the account ceiling it would have read 600 that day, and it read 2500.

## The SoC limits came from the rare source

The previous review found two sources writing the SoC limits and kept `254/39 f29`. The principle was right and the survivor was wrong. Over a thirteen-minute window with identical buffer retention, `32/2` carried the limits 83 times and `f29` four times, and a limit changed in the app reached `f29` four minutes and twenty-seven seconds after it reached `32/2`. Changing a limit in the app took minutes to show up instead of seconds.

The two never disagreed across four value pairs, so the flap that finding guarded against is not present on this device.

The startup path was the risk here and it is clear: the get-all bundle carries a `(32, 2)` block with both fields, decoded from `tests/fixtures/stream_ac5000/es22_get_reply_masked.json` rather than taken on trust, and the existing replay test can now only pass through that block. Our own compiled proto names field 7 `max_charge_soc` and field 21 `min_dsg_soc`, which is a second independent agreement.

## Review pass on top

- A comment introduced with the fix called both `f10.5` and `f10.6` account ceilings. Only `f10.6` behaves that way; `f10.5` went 600 to 800 when the ceiling rose and matches no app setting. In a change whose purpose is retiring overclaims, that one had to go too.
- The sensor comment still described the input limit as account-level. It is an ordinary app setting.
- The replay test still allowed a `32/2` frame to parse to nothing, which was correct while nothing in it was mapped and is now a hole. Removed, and a new test asserts every `32/2` frame in the capture yields both limits.

## Tests

2164 pass, up from 2162 on main. Three of his tests fail under the old mapping rather than merely passing, so they pin the new behaviour.

Ref #177


shuette42:
Superseded by #213.

PR #223, opened 2026-08-08
feat: add EcoFlow STREAM AC 5000 (ES22) controls

slaapyhoofd:
Adds controls for the EcoFlow STREAM AC 5000 (`ES22`): two switches, five numbers and a
work-mode select, all Enhanced only since the device is not reachable through the
Developer API.

This is the second half of #177, split from the read support as agreed. Every constraint
from that discussion is met: one readback source per entity with the reason recorded, no
control that cannot be confirmed from the device, Enhanced Mode only with no HTTP path
added.

## Every frame is the app's own

These are not inferred. The Android app publishes its writes to the same
`/app/{uid}/{sn}/thing/property/set` topic the integration subscribes to, so each builder
reproduces a captured app frame with the payload swapped. Three are asserted byte for byte
in `tests/test_stream_ac5000_commands.py` against the bytes the app sent.

Each write was then confirmed on live hardware three ways: the device acknowledged it, the
new value came back on the telemetry stream, and it was read back in the EcoFlow app
itself. The last one matters, because a device echoing a setting is not the same as the
setting having been applied.

Confirmed that way: work mode both directions, discharge power stepped 600, 300, 0, 600
with the battery physically settling where commanded, charge power on a unit with no
charge task (which creates one) and its removal, both SoC limits read back on three
independent fields, backup reserve at 55% then 30% matching the app, and the backup socket
toggled from Home Assistant and followed by both the app and the telemetry.

## A power setpoint is a scheduled task

This device has no direct power control. A scheduled task is its setpoint, so setting a
power writes a task covering the whole day, enabled, at that power.

That has consequences worth reviewing rather than discovering:

- **A power write replaces the schedule.** The task written always covers the whole day
  and is always enabled, so the value asked for is the value that acts. A setpoint that
  only applied inside a window somebody set in the app would not be a setpoint.
- **The charge task's SoC target is carried over**, because it decides what charging does
  rather than whether it happens. A task set to stop at 80% must not be reset to 100% by a
  change of power.
- **Charge and discharge are separate whole-day tasks, so setting one removes the other.**
  Two overlapping whole-day tasks are something the device acts on by doing nothing at
  all, and the app names it as such. A removal that fails is not followed by the write,
  since that is exactly how the overlapping pair would be created.
- **Zero is a real setpoint.** A 0 W discharge task parks the battery, while removing every
  task leaves it running its own 200 W base output.

## The setpoints are not limits

Worth stating plainly, because the naming invites the opposite reading. The app calls both
task powers a maximum, so the numbers sit beside the Max Grid-tied Output Power and Max
Grid Input Power *sensors*, which are the actual ceilings.

Neither ceiling is writable. The write surface is config fields 19, 25, 29, 30 and 39, and
none of them is a power limit, so both stay read-only sensors. A setpoint above a ceiling
is accepted, acknowledged and then clamped by the device rather than refused. Measured on
an external meter on the AC side: a standing 2500 W charge request delivered 2470 W with
the app's input limit at 2500, and 1218 W after the limit was moved to 1200, with the
transition one 5 s sample wide in both directions.

So nothing here can drive a unit past the limit its owner configured, and the entity
documentation says so rather than leaving it to be inferred.

## Whether a smart meter is linked changes what a setpoint means

The one thing worth knowing before automating this device, and it is an installation
choice rather than a model difference.

With a meter linked the device runs closed loop against it and will not discharge into an
export, so the setpoint acts only as a ceiling on covering house load: 1400 W requested
into a house needing 200 W delivers 200 W, and enabling feed-in does not change it. With
no meter linked it runs open loop, which the app's own help text describes as any unused
power flowing to the grid, and the setpoint becomes an absolute power command: unlinking a
Tibber Pulse turned that same 1400 W request into a measured 1400 W discharge.

An optimiser that wants to command power rather than cap it should therefore do its own
metering. Both cases are in the entity documentation.

## Readback sources, one per entity

| entity | source | why that one |
|---|---|---|
| max charge SoC, min discharge SoC | `32/2 f1.7` / `f1.21` | `f29` lags an app change by minutes; `f33.7`/`f33.8` are a third copy. `32/2` is the fast one |
| charge task power | `40.1.8.3.3` | the only field that followed a written value |
| discharge task power | `40.1.9.1` | tracked all eight steps of a driven ramp |
| charge target SoC | `40.1.8.3.2` | read back and preserved on every power write |
| work mode | `254/39 f25` | confirmed against the app on both transitions |
| backup reserve, backup socket | `f30.1`/`f30.2`, `f19.1` | confirmed against the app |

One asymmetry is now recorded in the code, because it is the kind of thing the next reader
will assume away: a task is written on config field 39 and read back on `f40`. The grammar
is identical on both sides and only the top-level number differs. Every other config field
reads back under the number it was written on.

## Failure modes

Two ways a write can be refused, told apart deliberately. A value the device will not
accept is reported as rejected, with a translated message rather than the traceback a bare
`ValueError` produces. A write needing a counterpart the device has not reported yet uses
the existing `raise_set_not_ready`, the same helper the Delta 3 port priority uses for the
same situation: it clears with the next status frame, and calling it unsupported would
send the owner looking for a device limitation that does not exist. Both paired settings
hit this, since field 29 carries the two SoC limits together and field 30 carries the
backup reserve flag with its level.

A setpoint is only acted on in custom mode. In the other modes the write is accepted and
ignored, and since the device reports no error, a warning goes to the log instead.

Deleting the last task in the app stops the readback rather than zeroing it, so both power
entities report unknown once the device says its task list is empty, instead of going on
showing a task that no longer exists.

## Testing

`pytest tests/` green, 2252 tests. New coverage in
`tests/test_stream_ac5000_commands.py` (byte-for-byte vectors) and
`tests/ha/test_stream_ac5000_controls.py`.

Beyond the suite, this branch has been driving my own unit continuously under an external
optimiser for about 48 hours: roughly 300 setpoint changes across the two power numbers,
194 of them distinct values, work mode held at custom throughout, no stuck task and no
write that failed to read back. The most recent check after deploying this exact commit:
1159 W commanded, 1159 W read back one second later, 1158 W measured at the battery.


shuette42:
Reviewed. Nothing blocking. Two things to fix, one question, then this goes in as beta.6, rebased so your authorship survives like it did for the read support.

Before the findings, what the evidence bought. 48 hours under EMHASS, roughly 300 setpoint changes, 194 of them distinct, no stuck task and no write that failed to read back. Nobody else on this device could have produced that, and it is the reason this review is about naming instead of about protocol. The protocol questions were already answered by your captures.

Three things I would have raised if you had not already handled them:

**You did not reuse the shared SoC number keys.** I checked what would have happened: the shared `min_discharge_soc` number displays "Discharge Limit", so reusing it would have put a second wording on one setting and forced a rename on every Delta 3 and PowerOcean owner. Your ES22-specific keys avoid that entirely.

**`f40.1.8.3.2` is read back and preserved on every power write**, with a test asserting 80 present and 100 absent in the frame. Without it, a task set to stop at 80 percent silently becomes one that charges to 100.

**The write is abandoned when the task removal fails** instead of proceeding into the overlapping pair the device responds to by doing nothing. That is the failure that would have been hardest to diagnose from a bug report.

And on the export ceiling: you were right that the entity I asked you to constrain does not exist, and you said so plainly rather than quietly not doing it. That is the more useful answer.

## Finding 1: the sibling setpoint latches after a cross-kind write

The one real finding, and both passes over the code found it independently.

Set a discharge power while a charge task exists. The charge task is correctly removed first, then the discharge task is written. But nothing clears `scheduled_charge_power_w` locally, and the device will not clear it either: a deleted task simply stops being mentioned, and the clear-to-unknown path only fires when the task list is empty, which it now never is because the new task exists.

So Max Grid Charging Power and Scheduled Charge Power both go on displaying, indefinitely, a task the integration itself deleted. An optimiser reading that entity reads a configuration that does not exist. Second order effect: every later discharge write sees the stale value and sends a removal frame for a task that is not there. Your soak shows the device tolerates that, so this is display honesty rather than device safety.

Suggestion, and it is a suggestion since you have the frames: after the removal frame succeeds, clear the removed kind's power, enabled and window keys locally, mirroring what the parser's empty-task block already does. **Keep `scheduled_charge_soc_target`.** Clearing that too would resurrect exactly the reset-to-100 bug the preservation logic exists to prevent. It also self-heals in the right direction: if a removal is acked but not applied, the device re-reports the surviving task and repopulates the key.

## Finding 2: the four control names differ from the sensors reporting the same value

Checked against `main` rather than assumed:

| value | sensor name today | number name in this PR |
|---|---|---|
| `scheduled_charge_power_w` | Scheduled Charge Power | Max Grid Charging Power |
| `scheduled_discharge_power_w` | Scheduled Discharge Power | Max Discharging Power |
| `max_charge_soc_pct` | Charge Limit | Max Charge SoC |
| `min_discharge_soc_pct` | Discharge Limit | Min Discharge SoC |

All four controls carry a different name from the sensor showing the same number, and the power pair has been published as "Scheduled ..." across five betas.

Three reasons this is more than tidiness, and the second is the one that decides it:

1. Home Assistant gives a `number` no description field, which is why the wording ended up in the docs. The name is the only text an owner sees while dragging the slider, so the name has to carry the whole warning.
2. "Max Discharging Power" sitting beside the Max Grid-tied Output Power and Max Grid Input Power sensors reads as a fourth limit, which is the misread we both already agreed is real. And with no meter linked it does the opposite of capping: 1400 W requested becomes 1400 W commanded, by your own measurement.
3. `entity_id` is derived at creation and permanent. Anyone installing beta.6 keeps whatever name ships, forever.

My position is that all four should match their sensors, so the powers become Scheduled Charge Power and Scheduled Discharge Power, and the SoC pair becomes Charge Limit and Discharge Limit. I would rather hear your view than decide it, because you have the app in front of you and I do not.

One genuine question that settles part of it: is "Max discharging power" verbatim from the app for the **discharge** task? You quoted the app for the charge task. For discharge we only have "the app calls both a maximum". If the app really uses that string, this is a discussion about which vocabulary wins. If it does not, it is a rename.

## Finding 3: minor

The two switches lack `enhanced_only=True` while your numbers and select carry it. No practical effect, since an ES22 cannot be configured with developer keys at all, but it is an inconsistency inside one entity set and the flag documents the intent for the next reader.

## Documentation

The entity page explains the schedule replacement mechanics thoroughly, but it never tells an owner who also uses the EcoFlow app's own scheduler not to mix the two. Everything needed to deduce it is on the page. Nobody deduces it while dragging a slider. One sentence is enough.

## One more thing

Config field 19, the backup socket, is in and welcome. My sentence on the 8th said the write surface was 25, 29, 30 and 39. That was my omission, not your addition: you announced both switches on the 6th.

Separately, and not for this PR: the four energy counters on this device were not accumulating, which I traced today. The cause is on our side, in how power readings are added up over time, and the fix is in progress. It does not touch anything you have written here.

slaapyhoofd:
All three are in, pushed as `6211c35` on the same branch. 2268 tests green, hassfest clean.

## Finding 1: fixed the way you suggested

Once the removal frame is acknowledged, the removed kind's `power_w`, `enabled`, `start_min` and `end_min` go to None locally, in both the coordinator's store and the snapshot entities read from, so the setpoint reports unknown at that moment instead of waiting for a readback that never comes. `scheduled_charge_soc_target` survives, for exactly the reason you gave: `_build_stream_ac5000_task` reads it back and falls back to 100, so clearing it would be the reset-to-100 bug arriving through a different door.

Only a successful removal clears. If the frame is not acknowledged the task may well still be there, so nothing is forgotten and the write is still abandoned. There is a test for each of those, and one for the second-order effect you described: two discharge writes in a row now send one frame the second time, not two.

Your reading of why it could never self-heal was right, and the empty-list branch is precisely where I had assumed the clean-up was already happening.

## Finding 2: the powers renamed, and a wrinkle on the SoC pair

The question first, since it is a fact rather than a preference: **yes, "Max discharging power" is verbatim from the app**, for the discharge task, next to "Max grid charging power". Both are where the current names came from.

I have renamed them to yours anyway, and the app string is the reason rather than the obstacle. The app's word is true closed loop only. With no meter linked the device delivers what it is asked for, 1400 W requested and 1400 W measured, and that is the configuration anyone driving this from an optimiser will be in. So "Max" is wrong in exactly the case where the control gets used, and "Scheduled ... Power" is true in both meter states. Your point 2 and plain accuracy land in the same place. Key and displayed name both changed, so it is `number.<device>_scheduled_charge_power` sitting beside the sensor of the same name.

The SoC pair I have left alone, and this one has a wrinkle worth knowing about for future reviews, because it caught me out too when I went to make the change. The `name` field on `EcoFlowSensorDef` and `EcoFlowNumberDef` in `const.py` is not what gets displayed, and it is not read anywhere in the integration. Both platforms set `_attr_translation_key = definition.key` with `has_entity_name = True` and neither sets `_attr_name`, so `translations/en.json` decides the name. What an owner sees today:

| value | sensor | number |
|---|---|---|
| `max_charge_soc_pct` | Max Charge SoC | Max Charge SoC |
| `min_discharge_soc_pct` | Min Discharge SoC | Min Discharge SoC |

"Charge Limit" and "Discharge Limit" live on the sensor definitions at `const.py:877-878`, so that is very likely where the table's third and fourth rows came from, but they never reach the UI. Renaming the pair to those would mean renaming `sensor.max_charge_soc_pct` and `sensor.min_discharge_soc_pct`, which PowerOcean and Delta 3 share, which is the same blast radius that kept me off the shared number keys to begin with.

Happy to make the whole set speak the app's vocabulary if you would rather, but it is a larger change than this PR and it lands on other people's devices, so it felt like yours to decide rather than mine to slip in.

There is now a test asserting that every ES22 number's translated name equals the translated name of the sensor for its state key, so the pair cannot drift apart again quietly.

## Finding 3: done

Both switches carry `enhanced_only=True`, and a test asserts it across every ES22 number, switch and select at once rather than for those two alone. Free change: the switches are new in this PR, so no installed base loses an entity when the flag starts gating them.

## Documentation

One sentence under the setpoint bullet: do not run the app's own scheduler alongside these controls, because each side overwrites the other.

The stale-task note needed splitting as well, since after finding 1 the two cases behave differently. A task deleted in the app still leaves its last values behind until one of that kind is reported again. A task this integration deletes clears immediately, because it knows what it deleted.

## The rest

Field 19 needs no correction on your side. It was in the 06-08 list either way, and an omission in a summary is not the same as something appearing unannounced.

And thank you for tracing the energy counters. I went looking at mine too, from a frozen `solar_energy`, and got as far as the gap branch before working out that mine was self-inflicted: I unlinked the Tibber Pulse four days ago and this device cannot report solar at all without a meter, so the counter had nothing left to integrate. Good to know the real cause was somewhere I would not have found from here.

shuette42:
You are right about the SoC pair and my table was wrong. I checked: both platforms set `_attr_translation_key` from the definition key with `has_entity_name` and neither sets `_attr_name`, so `translations/en.json` decides, and there both the sensor and the number read "Max Charge SoC" and "Min Discharge SoC". The `name` field on the definitions in `const.py` is not displayed anywhere. I read "Charge Limit" out of `const.py:877-878` and put it in a table as what an owner sees. It is not.

So rows three and four of that table describe a mismatch that does not exist, and you went and looked for it because I wrote it down. That one is mine.

Leaving the pair alone is right. Renaming it would move `sensor.max_charge_soc_pct` on PowerOcean and Delta 3 as well, which is the blast radius you avoided by not reusing the shared number keys in the first place. The test asserting every ES22 number's translated name equals its sensor's is the better answer than either of us renaming anything, because it holds for the next person too.

On the app vocabulary question: leave it. Making the whole set speak the app's wording is a real change and it lands on devices whose owners are not in this thread, so it wants its own pass rather than riding in here.

Thank you for answering the discharge label as a fact rather than as a preference, and then renaming anyway for a better reason than mine. "Max" being true only closed loop, and wrong in exactly the configuration an optimiser runs in, is the argument I should have made and did not have.

The correction in #177 on the task grammar is noted, including that it was your own table from the 3rd that was wrong rather than the frame. Three more copies of the same claim found and fixed is the part worth having: nothing executes a comment, so a wrong one survives every test run there is.

Finding 1, the removal clearing only on acknowledgement, with the target SoC deliberately surviving, and a test for the second-order case, is exactly it.

I will run this through the deploy check and merge it rebased, so it lands under your name like the read support did. One thing to know before it goes out: the four energy counters on this device were stalling for a reason on our side, which I traced today and fixed separately in #224. Your `solar_energy` was a different cause, as you found. Both will be in the same pre-release as this.

shuette42:
Second pass done. One blocker, two questions, two things that are ours rather than yours.

First what held up. I test-merged `6211c35` against current `main`: clean, zero conflicts, 2277 passing. I compared test IDs rather than trusting the total, so your 2268 are all there and accounted for. Your branch is one commit behind because #224 landed in the meantime; the two do not interact, your PR touches no integrator code. A rebase before merge, nothing more.

Your fix for finding 1 goes through `set_device_value`, the same seam the optimistic updates already use. That was the part I was most worried about and it is not a problem.

## Blocker: two concurrent power writes recreate the overlap the sequence exists to prevent

`_async_set_stream_ac5000_value` reads the sibling setpoint, awaits the removal, clears locally, then awaits the write. Two suspension points and no serialisation. Charge and discharge are separate entities, so Home Assistant runs their service calls as concurrent tasks. Interleaved:

1. A (charge) removes the discharge task
2. B (discharge) removes the charge task
3. A writes the charge task
4. B writes the discharge task

Both tasks exist again. The device reports overlapping time periods and acts on neither, while both entities report an optimistic success. The trigger is your own headline case: anything that sets both setpoints in one pass.

## The fix, and it settles a second thing at the same time

Move the whole remove-then-write sequence into `coordinator/set_commands.py` as a single coordinator method, rather than putting a lock in the entity.

The reason for that shape rather than a lock is the more interesting half. `_clear_stream_ac5000_task` is doing something none of the other optimistic writers do. The established invariant is that an entity may seed the store with a value it just sent, on the key it owns, and device truth overwrites it later. This clear writes keys the entity does not own, it writes a retraction rather than a value, and no device truth will ever overwrite it, because the parser's empty-container clear cannot fire while the replacement task keeps the list non-empty. The knowledge behind that retraction is real and only the integration has it. The question is where it gets asserted from.

The coordinator is the single place for data flow logic here and the entity classes stay thin. A two-frame transaction plus a state retraction is coordinator work, and there is precedent right next to it in the debounced PowerOcean SoC flush. Moving it down serialises the sequence, puts the retraction next to the store that owns it, and keeps `set_device_value` meaning "seed your own key". One change answers the race and the boundary.

## Two questions, not findings

**1. Does the device match a task removal by kind, or by kind and window?** The removal frame identifies its victim by kind plus a whole-day window at 0 W. A task created in the app carries the user's own window. If the device matches on the window too, the removal silently misses, the new whole-day task overlaps the app's, and the device acts on neither while Home Assistant shows success. One test on hardware answers it: create a charge task in the app with a window of its own, say 00:00 to 06:00, then set a discharge power from Home Assistant, and see what the app's schedule page shows.

**2. Was a charge power write ever observed arriving back on `40.1.8.3.3` from the device?** The discharge side tracked all eight steps of your ramp, which is unambiguous. For charge the readback rests on grammar symmetry with the write, and the parser note on `main` says the charge power was never seen to move. If you have a frame showing it follow a written value, that closes it. If not, it is worth saying so in the code rather than leaving symmetry to carry it.

## Two things to note, nothing to do

- A task deleted **in the app** while its sibling exists still latches in Home Assistant, because the wire has no per-task retraction and the empty-container clear cannot fire while the other task lives. That predates your controls. It is ours to decide, not yours to fix here.
- A frame already in flight when the removal is acknowledged can re-merge the removed task's keys and restore the latch behind a race. Narrow window, since task blocks ride in roughly 2.7 percent of frames. If the sequence moves into the coordinator, the suppression has a natural home there.

slaapyhoofd:
Pushed as `6046026` on the same branch.

Both questions are answered from hardware and the blocker is fixed. One new finding fell out of the test session; it is a read-path one, so I have deliberately kept it out of this PR.

## The blocker: fixed as you described, and it turned out to cover two more writes

The whole sequence now lives in `coordinator/set_commands.py` as `async_set_stream_ac5000_task_power`, under one `asyncio.Lock` held across the read and both frames. The retraction moved with it, so it sits next to the store it writes and `set_device_value` goes back to meaning "seed your own key". Your reasoning for that shape rather than a lock in the entity holds up better than I expected, for two reasons I only found by doing it.

**The same race is on two more config writes.** Field 29 holds both SoC limits and field 30 holds the backup reserve flag together with its level, so each of those is a read-modify-write too. Interleaved they do not produce an overlap, they lose one of the two changes: whichever frame lands second carries the other's value from before it was changed. Field 30 is the one that settles the design question, because its flag is a switch and its level is a number. `PARALLEL_UPDATES = 1` would have been a one-line fix for the task pair and for field 29, and it could never have covered field 30, since the platform semaphore is per platform. All three take the same lock now.

**The lock on its own is not sufficient.** The value a write sends has to be visible in the store before the lock is released. The entity applies its optimistic value after the coordinator returns, so a waiter acquiring in between reads the state from before the write it was queued behind, which is the same bug one step along. Each of the three now seeds what it sent, inside the lock, and only once the frame was delivered.

Tests: one per race. I checked they earn their place by rerunning them against a copy of the tree with the lock swapped for a two-permit semaphore, where all three fail. Two details in case you write one of these yourself: an `AsyncMock` completes without ever handing control back to the event loop, so a race test written against one passes whether or not the sequence is serialised, and the sender in these suspends instead. And the power test is parametrized over which of the two writes starts first, because only one of the two orders is harmful and nothing decides which one Home Assistant picks. It passed on the benign order before I noticed.

2281 passing.

## Question 1: matched on kind, not on window

Tested on hardware. I created a charge task in the app with a window of its own, 02:00 to 06:00 at 800 W, then set a discharge power from Home Assistant:

```
16:01:23  app write    add    .2=1  window=02:00-06:00  39.1.9.1=800
16:06:28  readback     .2=1  window=02:00-06:00
16:10:42  our write    remove .2=1  window=00:00-23:59  power=0
16:10:42  our write    add    .2=2  window=00:00-23:59  39.1.9.1=300
16:10:43  readback     .2=2  window=00:00-23:59  power=300
```

The 02:00 to 06:00 task never appears again, in the remainder of the capture or in Home Assistant. So the removal frame's whole-day window does not have to match the victim's, and the silent miss you described does not happen.

Worth recording for whoever reads this next: the app itself does not send a generic removal. It echoes the entire task back with operation 3, real window, power and target SoC included, which I had assumed meant the device matches on all of it. It does not, but the app's behaviour could not settle that either way. It had to be tested.

## Question 2: yes, observed following a written value

```
15:57:30  our write    update charge  power=598
15:57:32  readback     40.1.8.3.3=598  window=00:00-23:59  soc_target=100
```

Two seconds, and it is a value that moved rather than a static match: 691 W before, 598 W after. The window is a second fingerprint, since this integration writes 1439 as the end minute where the app writes 1380, and the readback carries 1439. So `40.1.8.3.3` is a real readback and not symmetry with the write.

I have corrected the parser note in this PR rather than leave symmetry carrying it. The old wording, that `f33.9` sitting at 600 left the scheduled charge power unconfirmed, conflated "which field is it" with "does it read back", and both halves are now answered.

## One new finding, and I think it belongs in its own PR

The app frame that created the task above is not what the parser assumes:

```
39.1.1 = 1 (add)     39.1.2 = 1
39.1.7 = 02:00-06:00
39.1.8.3.2 = 100     39.1.8.3.3 = 598    <- the charge power we had set, dragged along
39.1.9.1 = 800                            <- the discharge power just configured
```

One task object carrying both power blocks, added with the same `.2` value our whole-day charge task had, and it replaced ours rather than sitting beside it. The device echoed it back as `40.1.2 = 1` with only `40.1.9.1` present.

`_finalize_task` takes the kind from `.2`, so it read that as a charge task, looked for the power at `40.1.8.3.3`, found nothing, and set no key at all. For the nine minutes it stood, Home Assistant showed Scheduled Discharge Power as unknown while the device was holding an 800 W discharge task, and Scheduled Charge Power as 598 W for a task that no longer existed.

That lands underneath the controls in this PR, because the sibling check reads exactly those two keys. A task the parser cannot see is a task the write path will not remove, so it writes on top of it: your overlapping pair again, arriving through the read path with no concurrency involved.

I have not fixed it here, because the obvious fix has a trap in it. Taking the kind from which power block is present discriminates cleanly across every capture I have. But if `.2` is an identity rather than a type, then a correctly parsed discharge task in slot 1 would afterwards be removed with `.2 = 2` and the removal would miss, which is precisely the failure your question 1 was hunting, reintroduced by fixing the read. The safer shape is probably to carry the observed `.2` through into the removal frame instead of re-deriving it from the kind. I have exactly one instance of `.2 = 1` carrying a discharge power, so that is a hypothesis and not a conclusion.

Happy to open it against `main` as its own PR with the frames attached, or to leave it with you if you would rather look at it alongside the app-deleted-task latch, since both are the same seam.

## The two you listed as yours

Nothing from me on the app-deleted-task latch. On the in-flight re-merge, one number that might help size it: across the captures, a frame carrying a task block arrives every 24 to 52 seconds, so the only real exposure is a frame already in transit when the removal is acknowledged, well under a second. A suppression window of a few seconds would cover it, and anything longer would start swallowing a task the user genuinely creates in the app right after a write.

shuette42:
Merged and released. Rebased, so all three commits sit on `main` under your name and email. It is out as `v1.17.0-beta.6`.

https://github.com/shuette42/ecoflow-energy-ha/releases/tag/v1.17.0-beta.6

**What the fix round added beyond what it was sent.** You were given one race and came back with three. Field 29 and field 30 are the same read-modify-write pattern and neither was in the report. The second finding is the one that would have been missed by anyone implementing only what was asked: the lock on its own is not enough, because the written value has to be visible in the store before the lock is released, or a waiter acquiring in between reads state from before the write it was queued behind. A correct lock with the seeding left outside it looks finished and is not.

**Your `PARALLEL_UPDATES` reasoning holds.** I checked it: field 30's flag is a switch and its level is a number, so a per-platform semaphore could never have covered it. That is exactly why the one-line fix would have looked sufficient.

**The test verification.** Reproduced here independently against a two-permit semaphore: three fail, one passes, and the one that passes is the benign starting order. So the parametrization is load-bearing rather than decorative, which is the part that would have quietly rotted if it had been written the other way.

**Both hardware answers landed.** Removal matching on kind rather than on window is settled by a real app-created 02:00 to 06:00 task disappearing, which no amount of reading the app's own frames could have told us. And the charge power readback following a written value, 691 to 598 in two seconds, with the end minute as a second fingerprint. That second one is what let the parser note be corrected instead of leaving symmetry to carry it.

**The read-path finding: yes, please, its own PR against `main`, with the frames attached.** You were right to keep it out of this one. Your reading of why it matters is the reason it goes near the top: a task the parser cannot see is a task the write path will not remove, so it writes on top of it. That is the overlapping pair again, arriving through the read path with no concurrency in it at all.

Your instinct on the trap is right too. Carrying the observed `.2` through into the removal frame rather than re-deriving it from the kind is the safer shape, and one instance of `.2 = 1` carrying a discharge power is a hypothesis, not a conclusion. Say so in the PR and let the frames decide it.

The app-deleted-task latch stays with me. Nothing needed from you there.

Issue #231, opened 2026-08-09
Stream 5000 support


PSCourtney:
### Device Type

New Device Support

### Problem or Use Case

Add stream 5000 support

### Proposed Solution

like the AC 5000

### Additional Context

<img width="537" height="597" alt="Image" src="https://github.com/user-attachments/assets/559877b2-1ebb-46c8-8135-27b8f5fc001e" />

PSCourtney:
[config_entry-ecoflow_energy-01KZKCYFC33JHS8T145GN7BJ56 (1).json](https://github.com/user-attachments/files/30877345/config_entry-ecoflow_energy-01KZKCYFC33JHS8T145GN7BJ56.1.json)

Silly me, I wasn't using the beta version. 
Above is my diagnostics dump. 

Hope it's helpful?  I suspect the ES22 parser will work for ES21 but there will be a number of sensors for the MPPTs

shuette42:
Thanks for the raw capture, and for calling the ES22 parser as the likely fit. You were right, and having both the recording and the correct guess in the same post turned this into hours instead of days.

The STREAM 5000 (ES21) is now recognized and gets the same entity set as the STREAM AC 5000: 50 sensors and 2 binary sensors. It goes into the next pre-release of the current cycle.

On why the existing parser is the right one: EcoFlow's own product list has STREAM 5000 and STREAM AC 5000 as the same product in two models. Running your six frames through the existing parser gives numbers that add up against each other, 630 W leaving the battery arrives as 611 W in the house and 19 W to the grid.

What does not come with it is the controls. Every write command for this family is a rebuild of a frame captured from an AC 5000, and a power setpoint writes a scheduled task into the battery. Reading the same values back proves the device speaks the same language, it does not prove the same write is accepted. I am not guessing on something that writes to a battery.

Your MPPT hunch is open rather than settled. Your unit does send two blocks an AC 5000 does not, so there is something there. But there is exactly one frame of it and every value in it sits still, and you cannot name a field from a photograph.

That is where a longer recording would help. Switch the raw capture on and leave it running through a stretch where things actually move, sun coming and going, load changing. Note what you see and the time, then download diagnostics right after. That answers the MPPT question, and it is the same evidence that would eventually make the controls possible.

One side note from your device list: the SM2B in there is an EcoFlow x Shelly Smart Plug (UK) according to EcoFlow's product list, not a meter. It is not supported.

PSCourtney:
[...]

PR #232, opened 2026-08-09
PR to add Stream 5000 support. ES21

PSCourtney:
## Related Issue

#231  - TODO: add sensor support for MPPT's

## Summary

Adds the new device ID ES21. Support added from ES22 which is identical other than some missing sensors for the MPPT's.
Other than that they are functionally identical.


PSCourtney:
[config_entry-ecoflow_energy-01KZKCYFC33JHS8T145GN7BJ56 - fork.json](https://github.com/user-attachments/files/30879663/config_entry-ecoflow_energy-01KZKCYFC33JHS8T145GN7BJ56.-.fork.json)

Sensor dump - tested as working in my local HAS instance.

shuette42:
The naming point from this PR is in, at the one screen where it actually applied, and the changelog credits it back here. The controls stay out for the same reason they stay out on the AC 5000. Nothing left in here.

PR #233, opened 2026-08-09
fix: read the ES22 scheduled task list as a list

slaapyhoofd:
## Related Issue

Ref #177. This is the read-path finding I kept out of #223.

## Summary

It is three defects rather than one, all in `_finalize_task`, and the one I reported was the least of them. Each hides a live task from Home Assistant on its own, and a task the parser cannot see is a task the write path will not remove, so it writes on top of it: the overlapping pair the device answers by doing nothing, arriving through the read path with no concurrency in it.

Measured on hardware. On the evening of 2026-08-08 an optimiser parked my unit by writing a 0 W charge task. That task vanished from the readback, every later discharge write therefore skipped the removal, and the device sat at its own 200 W base output for 39 minutes while Home Assistant reported success on all of them.

I also have to correct my own description in #223. I said the 16:01 app frame was "one task object carrying both power blocks". It was not, and the reason I read it that way is the same flattening the parser does: it was two tasks in one frame.

## The three, each with the frame

**1. `40.1` is a repeated field and the walk flattened it.** The docstring says "A frame carries a single task and the device cycles through them". That is false: 16 frames across my captures carry two tasks. 2026-08-03 22:21:59:

```
task[1/2]  op=2  .2=1  en=1  win=03:00-17:00  chgW=1800  chgSoc=100
task[2/2]  op=1  .2=2  en=1  win=18:00-02:00  disW=1400
```

Production parser on that frame, before this PR:

```
{'scheduled_discharge_power_w': 1400, 'scheduled_discharge_enabled': True,
 'scheduled_discharge_start_min': 1080, 'scheduled_discharge_end_min': 120}
```

The 1800 W charge task is gone. Each block is now collected raw and decoded on its own, with **its own set of seen groups**. That second part is load-bearing: `seen_groups` is global to a walk, so splitting the results while sharing the set would mark both `40.1.8` and `40.1.9` seen and read every two-task list as one task doing both.

**2. `.2` is the task's number in the app's list, not what the task does.** Across my captures it reads 1 on 75 charge tasks and on 5 discharge ones, and 2 on 52 discharge tasks. The 5 come from the 16:01 frame, which was:

```
task[1/2]  op=3 (remove)  .2=1  win=00:00-23:59  chgW=598
task[2/2]  op=1 (add)     .2=1  win=02:00-06:00  disW=800
```

The app removed the task numbered 1 and reused 1 for a discharge task. On 08-03 it had 1 charge and 2 discharge. The kind now comes from which power container the task carries, which cannot be ambiguous.

**3. A task at 0 W omits its power field.** This is the one that cost the evening, and I had not seen it when I wrote the report. From 19:52 on 08-08, both tasks whole-day, device delivering nothing:

```
task[1]  .2=1  .3=1  .7=00:00-23:59
         .8.1=1  .8.3.1=<serial>  .8.3.2=100
         .8.3.3 ABSENT              <- power is 0, so it is not on the wire
task[2]  .2=2  .3=1  .7=00:00-23:59  .9.1=297
```

So the kind cannot key off `40.1.8.3.3`; it has to key off the container, and the power inside a present container has to be filled. That is reading proto3, the same argument `40.1.3` already rested on in this file, and it is now in a task-level fill table beside it. The discharge side has been seen sending its zero explicitly, so there the fill is a `setdefault` no-op. Only the charge side omits it.

Link 3 alone reproduces the incident. Fixing 1 and 2 without it still leaves the parked task with no power, and the write path removes the sibling only when its power is a known number.

## Naming the task on the way out

Re-deriving `.2` from the kind is what makes a frame miss when the app has reused a number, so a removal now carries the number the device reported, falling back to the derived one when nothing has been observed. A number is never seeded after a write: the device decides where a new task lands and a guess would put a fabricated one into the next removal.

**The update frame carries it too, and you only signed off on the removal.** The same argument reaches it, and the read fix is what makes that path reachable: an app-created task now populates the power key, which is what selects `TASK_UPDATE` over `TASK_ADD`. Happy to split it back out if you would rather look at it separately.

Under both readings of `.2`: if it is a number, this is the fix. If it is a type, the observed value equals the derived one in every frame except one that contradicts its own container, which is self-contradictory under that reading anyway. And a miss is now recoverable either way, because the read path can finally see the task it missed. Two mirrored instances is still an inference, not a conclusion. The read path never consults `.2` and the write path never interprets it, so neither reading changes what this code does.

If you want it settled: create two discharge tasks in the app with different windows and read their `.2`. Same number means a type, different means a number. Not needed to merge.

## Held back deliberately

With the list in hand, retracting a kind that is absent from it costs about six lines and fixes the app-deleted-task latch. It is not here. You said that one stays with you, and there are two reasons beyond that. A clear writes `None`, and `as_known_int(None) is None`, so an over-eager clear stops the removal being sent, which is the same door this PR closes. And it needs "a frame carrying `f40` is the complete list", which holds in every capture but which nothing else here has to assume.

## Known limitation, now visible instead of silent

The device can hold more than one task of a kind. These keys hold the last of each and a write replaces one, so a second same-kind task survives it. Logged at debug rather than modelled.

## Verification

- **2354 tests pass.** 21 of them fail against a tree with the three source files reverted, so they discriminate rather than decorate. The one test I first wrote that passed both ways seeded the coordinator by hand; it now drives the real frame through the parser instead.
- **Two real frames added as a fixture** (`es22_task_frames_masked.json`), the 08-03 two-task list and the 08-08 parked pair, replayed through the production entry point. Serials replaced by a run of X of the same length, and that fixture is now in the loop of the existing no-identifier guard. The 22:21 frame carries a second, ES21-prefixed serial alongside the ES22 one; both are masked.
- **Confirmed on hardware after deploying this branch.** Wrote charge power 0, and the device sent back a task with `40.1.8` present and no `40.1.8.3.3`, which now reads `scheduled_charge_power_w = 0` where before it set no power key at all. Then wrote a discharge power: the removal went out, the charge setpoint cleared, and the next task list carried the discharge task alone. Run at 10% SoC so nothing moved.

## Changes

- `stream_ac5000_proto.py`: repeated-group collection in `_walk`, per-task decode with its own seen set, kind from the power container, task-level zero fill, the observed number published per kind, docstrings corrected where they described one task per frame.
- `stream_ac5000_commands.py`: `build_task_payload` takes an optional `task_slot`; `None` reproduces today's bytes exactly, so the three app vectors are untouched.
- `set_commands.py`: removal and update name the observed number, `_clear_stream_ac5000_task` clears it.
- Tests, fixture, changelog, entity docs, version bump.

No entity, translation or count changes. `test_documented_totals` is still 50. ES22 only; nothing here is reachable from PowerOcean, Delta 3 or the BK-series Stream.

One behaviour change worth knowing for existing installs: a charge setpoint that read unknown for a parked task now reads 0.

## Checklist

- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated
- [x] Version bumped in `manifest.json`
- [x] README updated (not needed, no sensor or count change; entity docs updated instead)


shuette42:
Third one in a row, and this is the strongest of them.

Reading `40.1` as the repeated field it actually is, instead of keeping the last block, is the right call and it matches what I see in the frames. Taking the kind from the power container rather than from `.2` is right for the same reason, and the per-task seen set is the detail most people would have missed. That you measured all of it on your own hardware, including the 0 W parked task that started the whole thing, is what makes it possible for me to take a change this deep into the parser for a device I don't own.

The TASK_UPDATE part stays in, no need to split it out. Your reasoning holds: the read fix is what makes an app-created task carry a power key at all, so that path only becomes reachable through this change. It belongs in the same PR.

One thing has to change before this can go in. In `tests/fixtures/stream_ac5000/es22_task_frames_masked.json` there is a 36-character device UUID sitting in clear text inside the hex payload. Both serials are masked exactly as you describe in the PR body, so this is the one field that slipped through. The two existing ES22 fixtures mask it with 36 X, that's the pattern to match.

It needs to come out of the branch history as well, not just the current state. I merge your PRs with rebase rather than squash, so the commits land on main as they are and the old blob would come with them. The no-identifier check never looked at that field in the first place, which is why nothing caught it. I'll extend the check on my side.

Two things I'd rather ask than assert, since you have the device and I don't.

The kind now comes off the `40.1.8` container, but the zero fill for charge power hangs off `40.1.8.3`. What happens with a task that has `.8` and no `.8.3` at all? From here that looks like it would give the kind back without a power, which is the state this PR exists to remove. Have you seen a frame like that, or is `.8.3` always present once `.8` is?

Second, in `set_commands.py` the `task_slot` on the update comes from the state before the removal. Does your device renumber the remaining tasks when one is deleted? If it compacts them, the update would be naming a slot that is now empty.

Low priority, just so it sits somewhere: moving the `40.1` decode into the per-task pass means a decode error inside a task no longer rejects the payload. The guard for XOR payload candidates leaned on exactly that. It's latent, since ES22 never sets `enc_type`, so there's nothing to do about it now.

One more thing, and it's not on you: you bumped to 1.17.1, but 1.17.0 isn't fully out yet. That isn't visible anywhere public, so there was no way to know. The correction is small, manifest back to 1.17.0 and your changelog entry moved into the existing 1.17.0 block instead of getting its own. I have it ready and tested here. I would have pushed it straight onto your branch, but that isn't possible for forks owned by an organisation, so either I send you the commit or you do the one-liner yourself, whichever suits you.

The retraction you held back, the one that fixes the app-deleted-task latch: I've opened #234 so the work doesn't get lost. If you want to take it, it's yours, but there's no expectation attached. You've put more than enough into this one already.

Once the fixture is sorted, this goes into the next test build of the current cycle.

slaapyhoofd:
Sorry about the UUID. It is out, and out of the history: rebased onto current
main and force-pushed rather than masked in a commit on top. The fixture now
carries 36 X where the meter UUID was, matching the other two, the note says so,
and both frames parse to exactly what they did before. Your check missed it
because it looks for `[A-Z0-9]{15,}` and this one is lowercase with hyphens.
Leaving the extension to you as you said.

Manifest is back to 1.17.0 and the entry has moved into the existing 1.17.0
Fixed block, so no need to send the commit.

One conflict on the rebase: `_seed_stream_ac5000` is `_seed_device_values` since
#251. Resolved onto the new name, keeping the note about never seeding a task's
own number, reworded so it reads as an AC 5000 point rather than a rule for
every caller of a now shared helper.

**`.8` with no `.8.3`.** Never seen. I scanned every task block I have, the three
fixtures plus captures taken today: 19 blocks, 14 carrying `.8`, and all 14 carry
`.8.3` too. One of those 14 is the parked task and it looks the way the fill
expects, `.8.3` present with the serial and the target SoC, `.3.3` absent.

Your reading of the code is right all the same. What makes the two sides differ
is only nesting: discharge power sits directly under its kind container at
`40.1.9.1`, so that fill already keys off the container that decides the kind and
cannot come apart. Charge power is one level deeper. Keying the charge fill off
`40.1.8` instead of `40.1.8.3` would make them identical and changes nothing for
any frame observed so far. Happy to put that in here or as a follow-up, your
call.

**Renumbering on delete.** Not observed, and not what that number looks like from
here: `.2` is supplied by whoever writes the task and echoed back unchanged, so
there is nothing for the device to renumber. Nothing has ever been seen to change
a task's number while the task stood, and across those same 19 blocks charge
always reads 1 and discharge always reads 2, through today's adds and removals as
well.

The limit of that: most of it is our own writes, which pick 1 and 2 by kind, and
the case that would actually prove compaction is not in the corpus, two tasks
standing with numbers the app chose, one deleted, the survivor read back. I can
run that on the unit if you want it settled.

For what it is worth on the write side, the removal that precedes an update
targets the other kind, and the number on the update is our own kind's, read back
from the device. A compaction would have to renumber a task that was not the one
deleted, between that read and the write, inside the same lock.

The XOR guard point is fair and I have nothing to add to it.

Unrelated to the review: EcoFlow released firmware 1.1.4.44 on the 13th. I took
my unit from 1.1.4.35 to it and compared the full state before and after. Nothing
moved, `f40` included, so the fixtures here still describe what is in the field.

shuette42:
You don't need to run the two-task experiment. Its mirror was already sitting in the repo: a diagnostics download from an ES21 owner, taken on the 16th, carries a frame with two tasks standing at once, both of them charge, `.2` reading 1 and 2. That settles it. `.2` is a slot number, not a kind, and no app session on your unit is needed to say so.

The same frame says something about what this PR actually is. Run through the parser on main it reported a scheduled discharge, enabled, 11:00 to 12:00, with no power at all. Run through yours it reports what it is: a charge task, 660 to 720, 2500 W, SoC target 100, slot 2. So this is not only an ES22 write path problem. The read path was publishing a charge task as a discharge schedule on an ES21 today, on hardware that has no controls at all and therefore no way for the owner to notice by acting on it. You corrected something we ship, not only something you found on your own unit.

Defect 1 lands the same way, independently of your device. That frame carries two tasks where the parser's docstring said one, so the one-task-per-frame assumption was wrong on both models of the family, not just on the ES22.

Your commit is in main, with your name on it. I did not merge this branch: it was based on `2bf9c70` and four files had moved under it since, so rather than sending it back to you for a rebase I cherry-picked your commit onto current main, resolved the conflicts and pushed it through #261. The commit is `afac52b`, authored by you, unchanged apart from the conflict hunks. This PR is closed because the work is in, not because it was turned down.

One conflict was more than mechanical and you should know how it came out, because it touches your change. `_ZERO_FILL_PATHS` gained `12.2` and five `50.1.*` entries on main, the PV strings and the fix for their overnight latch, and your side removes `40.1.3`. Both survive. Your removal is right, and for a reason beyond the one you gave: `_walk` marks `40.1` as seen before the repeated-collect branch, so a frame-level fill would now deposit `_task_enabled_raw` in the parser output and nothing pops it again. The disabled-task reading itself is untouched by that, since it moved into your per-task fill. I checked both sides afterwards against that same ES21 capture: the charge task reads correctly and the PV strings still go to 0 W overnight instead of holding the last daylight value.

Compaction I'll take as a follow-up rather than a condition, and here is the reasoning so you can push back on it if you see it differently. The slot on an update is a number the device reported back for that same kind, the removal that precedes it targets the other kind, both sit inside one lock, and the number is never seeded from our own write, so a wrong slot corrects itself on the next readback. If the device does compact, the failure mode is a mis-aimed update, which is still better than the silent no-removal we had.

Two follow-ups worth doing, neither blocking. Your own suggestion of keying the charge zero fill off `40.1.8` instead of `40.1.8.3` should happen, since a task with `.8` and no `.8.3` yields a charge with no power, which is exactly the state this fix exists to remove. It needs a path-to-group indirection, so it wasn't part of this. The other is a comment tying the `_decode_task` swallow to `_pdata_candidates`, so the next person reading it sees that a per-task catch softens the guard the XOR candidate check leans on.

The identifier check was mine to fix and it went in alongside your commit. You were right that the old one could not match a lowercase dashed UUID. There is one guard now instead of two partial ones, it walks the fixture directory rather than a hand-written list, it covers the raw byte fixture that was outside both globs, and it scans a file as text rather than skipping it when it cannot read it as frames.

Thanks for the firmware comparison. 1.1.4.35 to 1.1.4.44 with a full before and after, nothing moved, `f40` included, is what keeps the fixtures worth anything.

PR #251, opened 2026-08-11
feat(stream): add AC Pro charge limit controls

Hendrixx-hue:
## Related Issue

Adds feature to #98 

## Summary

Adds hardware-verified charge and discharge limit controls for the EcoFlow Stream AC Pro (`BK31`) in Enhanced Mode.

## Changes

- Add Charge Limit and Discharge Limit number entities for Stream AC Pro.
- Restrict the controls to the hardware-confirmed `BK31` serial prefix.
- Reproduce the app's grouped ConfigWrite fields `6`, `33`, `34`, and `102`, including the required `from="ios"` header.
- Preserve all companion settings reported by the device instead of substituting defaults.
- Refuse writes until charge limit, discharge limit, and backup reserve have all been reported.
- Keep Backup Reserve at least three percentage points above Discharge Limit when raising the limit.
- Preserve Backup Reserve when lowering Discharge Limit.
- Serialize concurrent limit and reserve writes to prevent stale companion values from overwriting newer changes.
- Add English and German translations, documentation, and focused protocol/entity tests.
- Bump the integration version to `1.18.0`.

Both Home Assistant controls were tested on a live Stream AC Pro. A baseline of 95% Charge Limit, 20% Discharge Limit, and 23% Backup Reserve matched the EcoFlow app and device telemetry. Raising Discharge Limit to 21% moved Backup Reserve to 24%; lowering the limit again left Backup Reserve at 24%.

## Checklist

- [ ] Tests pass (`pytest`) — 2408 tests pass; one pre-existing, unrelated PowerOcean teardown test reports a lingering timer
- [x] CHANGELOG.md updated

Transparency: GPT 5.6 sol was used.


shuette42:
Thanks, this is merged. Charge and discharge limit are now Number entities for the Stream AC Pro in Enhanced Mode, written as one grouped write across the four fields. I pushed an extra commit onto your branch rather than sending you through another round.

Two things in your implementation are worth calling out. The read path takes its value from the telemetry stream instead of echoing the ack, which is the source that actually reflects device state. And concurrent writes are serialized with a real concurrency test behind it, not just the happy path.

The one substantial change is the clamping. Your note in #98 is the relevant evidence here: "there is a upper (?) / lower limit, which the unit accepts depending on min / max charge discharge % (-3%). Setting the backup reserve lower than this soft cap has no effect on the unit." That is the device enforcing a lower bound on its own, and it is exactly why the integration does not need to reproduce it. So two things came out: the hard rejection of an invalid combination, and the client-side pre-computation of the reserve value before sending.

On the rejection: what you described in #98 is a unit that ignores a value sitting too low, not one that refuses the frame. Raising an error in the client models different behavior than the one you observed.

On the pre-computation: it made the check inside this PR inconclusive. The client sent the 24 itself and the device reported it back, so that single run cannot confirm the rule. Not a complaint about the work, just the reason your #98 observation carries the point and this run does not.

What settled it was a side effect. Backup reserve has its own control. Anyone setting it below the discharge limit would have made the charge limit permanently unwritable, with an error pointing at a field they never touched. Now we send the reported values unchanged, the device clamps them itself, and telemetry reads back what actually applies.

Three smaller hardenings rode along: the prefix allowlist no longer matches on an empty serial number, the prefix constant is shared instead of one per feature, and the lock and its seed helper are shared as well. The `from="ios"` field is carried along too.

The version went back to the target of the current beta cycle, since that cycle has no full release yet. That is bookkeeping, not a judgment on your work. I also added three test cases to complete your suite.

Two things from my side. The min/max measurement from the #98 thread is still open, and that is what pins the bound down. Once it exists the integration can show the limit instead of guessing at it.

And #250 needs a rebase on the new main. One open question there: your outlet writes use two fields that the Stream telemetry parser already reads as PV voltage and PV current. Please cross-check that against your hardware. That kind of mapping has been wrong in this integration before.

PR #252, opened 2026-08-12
Added solar_pv sensors for ES21

PSCourtney:
## Related Issue

Adds to #231 

## Summary

Adds: 
- Missing solar PV sensors to ES21
- translation files added

image

**Field ids**
50.1.9 = PV1 
50.1.10 = PV2
50.1.11= PV3 
50.1.12 = PV4
50.1.3 = total PV 

-

## Checklist

- [x] Tests pass (`pytest`)
- [x] CHANGELOG.md updated
- [x] README updated (if new sensors, features, or behaviour changes)


PSCourtney:
Aside: 

FYI the `solar_w` field that was removed - should be **re-added.**  but renamed to `other_solar_w `

It represents "third party" solar in the ecoflow app.  Both the AC5000 and 5000 can report it. 

image

STREAM 5000: 
**Total solar = MPPT 1 + MPPT2 + MPP3 + MPPT 4 + Other** 

STREAM AC 5000:
**Total solar = other**

shuette42:
Thanks for staying on this.

The PR makes claims I can check: decimal-precise matches against app readings, fourteen frames across four captures on separate days. Those captures are not attached here and I cannot find them linked in the thread the way your earlier diagnostics dumps were. Every other field on this device got confirmed from the frames, including the naming point you contributed, so I need them before this merges.

To be clear about what this is: not "your numbers are wrong", but "I cannot tell yet without the frames". Attach the diagnostics files the mapping rests on, same as you did the first time, and I can work through them field by field.

On your aside: re-adding `solar_w` as `other_solar_w`, with total solar as MPPT 1 to 4 plus third party on the ES21 and third party only on the AC 5000, is a separate claim with its own evidence bar. Open, not accepted. It needs the same kind of anchor, an app screenshot showing that breakdown next to a diagnostics capture taken at the same moment.

PSCourtney:
[config_entry-ecoflow_energy-01KZN9ZATW4Z9JJFBNFM2F78DT (9).json](https://github.com/user-attachments/files/31113508/config_entry-ecoflow_energy-01KZN9ZATW4Z9JJFBNFM2F78DT.9.json)

image

Attached diagnostic. 


Total solar 265w
pv1 0
pv2 34
pv3 50
pv4 102

other 80

shuette42:
Thanks for the capture. That settles PV2, PV3, PV4 and the total. The total equals the sum of the four strings exactly in all seven frames that carry it, and your app reading of 34, 50 and 102 W is bracketed by the two recorded frames either side of that moment, 36.57 / 55.32 / 104.02 on one side and 27.39 / 49.02 / 98.07 on the other. The arithmetic checks out.

PV1 is the one still open. Every frame you have sent shows string 1 idle, so its position rests on the order of the fields rather than on a reading of its own. That is not a doubt about your numbers, it is a gap in what has been observed. If you get a capture while string 1 is actually producing, plus what the app shows at that moment, that closes it. No pressure, whenever it suits you.

I am taking the remaining work from here. The branch is rebased against the last week of changes, and I fixed a freeze where the readings would have held their last daylight value overnight. Two frames in your own recording, one full state and one update, carry the block with neither the total nor a single string in it. The existing Solar Power reading had the same problem a few weeks back, same fix. You should not have to chase a moving main twice in a row. It goes out in the current cycle, and credit carries over.

Your `other_solar_w` point holds up in the recording as well, 81 then 78 against the 80 you noted, and the strings plus that figure come to the 265 your app showed as the total. Open rather than accepted though, because it needs its own anchor the same way this one did. Tracked as #258.

One more thing your capture turned up, which has nothing to do with your PR. The flow carrying your own strings into the battery was never read at all, so Battery Power on a unit with PV on the EcoFlow was missing the whole solar charge. Two of your frames show the pack taking 34 W and 28 W and the reading was 0. That is fixed too, and it would not have been found without your recording.

shuette42:
Merged from #256, which carries your two commits with your name on them. The PV strings and the battery fix are both on main now. #231 stays open for string 1 and the blocks that are still unnamed.

PR #262, opened 2026-08-17
feat(stream): add AC Pro LED brightness control

Hendrixx-hue:
## Related Issue

Adds the LED brightness control requested in #98.

## Summary

Adds a writable LED Brightness number for the Stream AC Pro (`BK31`) in Enhanced Mode. The command path was captured from the EcoFlow app and verified on real hardware using HomeAssistant and comparing the value to the EcoFlow app.

## Changes

- Add an LED Brightness number with a range of 0–100% in 5% steps.
- Reproduce ConfigWrite field `384` with the required `from="ios"` header.
- Restrict the control to the hardware-verified Stream AC Pro (`BK31`).
- Use live telemetry field `994` as the authoritative state.
- Preserve read-only LED telemetry for other Stream models.
- Add payload, entity, model-gating, Enhanced Mode, and failure-handling tests.
- Update the changelog, version, README, and entity documentation.

## Checklist

- [ ] Tests pass (`pytest`) — 2,497 tests pass locally; one unrelated PowerOcean teardown error is also reproducible on unchanged `main` under Python 3.14.
- [x] `CHANGELOG.md` updated
- [x] Version bumped in `manifest.json`
- [x] README updated

Transparency: Codex GPT 5.6 sol was used to create the sourcecode.


shuette42:
Merged as #265, by rebase, so the commit sits on main under your name.

First, the thing I got wrong: I closed #98 this morning saying nothing further was asked of you, while this PR had already been open for hours and says in its own first line that it came out of that issue. That reads as if I had not seen it. I had not, and that is on me.

The code needed no changes. What stood out most is the test file: I changed field 384 to 385, changed the header source, put a wrong scale on the value, wrapped the frame differently, dropped the model gate and rescaled on the entity side, and every one of those broke between two and five of your tests. The strongest is the hand-derived literal frame, because it does not share an encoder with the code it checks, so it cannot agree with a bug by construction. I also checked the read-back claim against the parser rather than against your description, and it holds: 994 mapped, 384 deliberately left unmapped so the live value cannot jump to the slider target on an acknowledgement. And you reused the existing model gate and updated its comment in the same edit instead of adding a second one, which is the part contributors usually miss.

Two things I changed after review, neither a fault in your code. Your write went out straight from the entity, which made it the only Stream config write outside the shared device config lock, while that lock's own comment claims every Stream write goes through it. It now runs through a coordinator method beside its siblings. And the range guard in the builder now turns into a proper rejection instead of surfacing as a raw error, matching the other limit controls. Both got tests.

Two things outside the code as well. You set the version to 1.18.0, which is a fair reading of the rule that a running beta cycle takes no new features. I pulled it into the current cycle instead, so this ships in **v1.17.0-beta.21**. And the changelog entry named a tool from my local setup that does not belong in public files, so I rewrote it for readers.

Only thing worth doing: when beta.21 lands, move the slider once and check the brightness follows. The branch moved underneath your test, so it is worth one look.

shuette42:
v1.17.0-beta.21 is out on the HACS beta channel. Move the LED Brightness slider once when you have it and check the brightness follows.

Issue #247, opened 2026-08-11
Support for PowerPulse wallbox (serial prefix C376)


shuette42:
### Device Type

New Device Support

### Problem or Use Case

Reported by @dannywond in #88 (closed, PowerOcean Plus tracking), with earlier prefix work in Discussion #62.

The device is a PowerPulse wallbox with serial prefix `C376`, running alongside a PowerOcean Plus (`R371` family) on the same EcoFlow account.

### Proposed Solution

Read support for `C376`. Right now the prefix is all there is, no data of any kind.

`C376` does not match the PowerPulse 1 (`AC31`, tracked in #245), and it does not match the unconfirmed PowerPulse 2 mentioned in #7 either, so it gets its own thread rather than being folded into one of those.

### Additional Context

**Needed before anything can be mapped:** a raw data capture plus a diagnostics download from a `C376` owner.

1. Integration options, enable **Send raw data of unsupported devices to diagnostics (off after 24 hours)**
2. Leave it running, then download the diagnostics and attach the file here

The capture switches itself off 24 hours after being enabled. The download has to happen inside that window, otherwise the file will be empty and the capture has to be started again.

shuette42:
@dannywond an update here, and a correction to what I asked you for.

I asked you to switch on the raw data capture for the wallbox. That was the wrong request, and it would not have worked. A PowerPulse coupled to a PowerOcean says almost nothing on its own channel, so a capture of the wallbox itself would have come back nearly empty. The charging data travels on the PowerOcean instead, and the raw capture is only offered for devices that have no support at all, which your PowerOcean Plus is not.

A recording from another owner with a PowerPulse 1 settled where the data actually sits, and the readings are now built: charging power, session energy, session duration, charging state and the recognized vehicle. They go into the next pre-release of the current cycle.

Which turns your part into something much smaller. Once that pre-release is out, install it, and if your wallbox is coupled to your PowerOcean, look at the PowerOcean device for five new entities beginning with Wallbox. If they appear and hold sensible values while a car is charging, your `C376` reports the same way as the older model and there is nothing left to do here. If they stay absent during an actual charging session, that is the useful answer, and then I will come back with a request that fits your setup rather than the one I got wrong.

No need to switch anything on beforehand this time.

Xygen:
I have also a C376 and it is a PowerPulse 2 11kW.

shuette42:
[...]

PR #308, opened 2026-08-26
feat: read PowerGlow telemetry in Enhanced Mode

Xygen:
## Summary

Adds read-only Enhanced Mode telemetry for a PowerGlow coupled to a PowerOcean.

- Reuses the existing PowerOcean account-login MQTT connection and coordinator.
- Decodes the already-received PowerGlow reports: `212/8` water temperature (field 6) and `212/33` heating power (nested field 2).
- Treats a serial-only `212/33` proto3 item as the report's explicit idle value of `0 W`.
- Keeps PowerGlow an accessory of the PowerOcean: no new device type, serial-prefix detection, coordinator, login, poller, or transport.
- Adds no controls.
- Leaves target power and target temperature on the existing Standard Mode quota path. The observed Enhanced reports do not contain confirmed fields for either setting.

The existing PowerOcean accessory gating and disabled-by-default sensor definitions are reused; no entity definitions or translations are added.

## Evidence and scope

On the reporter's account-login PowerOcean, the integration receives both report types without the EcoFlow app being used:

- `212/8` contains a `77.0` water-temperature float, matching the separately observed PowerGlow reading of `77 °C`.
- Repeated idle `212/33` items contain the PowerGlow serial and omit the proto3 power field; the separately observed heating power is `0 W`.

A raw non-zero `212/33` value paired with the EcoFlow app's displayed heating power will be added before merge, as requested in Issue #7. The missing active-heating observation does not change the read-only scope or the existing connection architecture.

## Tests

- Hand-built protobuf fixtures use dummy `HF33TEST...` serials only.
- Regression coverage includes a bundled `212/8` + `212/33` frame and the explicit `0 W` idle form.
- `pytest tests/test_powerocean_bundled_frames.py -q` passes locally (28 passed).

The full Home Assistant suite is not runnable in this local environment because the available bundled Python is 3.14 while this repository's declared test dependency is pinned for Python 3.13; CI runs the project's Python 3.13 test matrix.

Ref #7


Xygen:
## Post-installation evidence: account-login idle run

I installed this branch on the reporter's PowerOcean with EcoFlow account sign-in and captured a new sanitized integration diagnostic. The run was healthy: WSS MQTT was connected and receiving for 469 seconds.

The coordinator snapshot contains both newly decoded keys:

- `heating_rod_water_temp_c`
- `heating_rod_power_w`

The live Home Assistant states were **76 °C** for Heating Rod Water Temperature and **0 W** for Heating Rod Power. The capture saw `212/33` 85 times during the run; its incremental idle frames decode as the serial-only proto3 form and now produce the explicit 0 W value. The bundled Get-All reply also includes the `212/8` parameter report and produces the water-temperature key.

The buffer tank is already above its target temperature, so the device does not accept a manually requested heating power. I will not force a heating cycle just to obtain a non-zero sample. Therefore this evidence confirms the existing-stream architecture, entity gating and idle decoding, but it does **not** claim a non-zero power-scale proof. The latter remains the only outstanding pre-merge evidence from Issue #7.

shuette42:
@Xygen The non-zero heating sample this PR lists as its outstanding pre-merge evidence was already in the diagnostics file you attached to #7 on 24 August. It is not in `212/33`, it is in `212/8`, as a plain integer on field 4: 2159 W drawn against 2160 W target, at 69 of 80 degrees.

That is the concrete reason the shipped version reads a different field. Across all seven `212/33` occurrences captured so far, in your file and in a second owner's PowerOcean recorded on #247, that report stays in the serial-only idle form. That includes the window where `212/8` reports the 2159 W draw. A power sensor bound to `212/33` reads 0 W straight through an actual heating cycle. `212/8` carries heating power, target power, water temperature and target temperature together in one message, so the decoder reads all four from there instead.

None of this was reachable from your side without the second capture, and none of it existed before your work. Your sanitised excerpt of the heating rod parameter report and the energy stream report gave the field names, you established the identity of the heating power field, and your idle run confirmed that the reports arrive unsolicited on the connection that is already open. That is what made the `212/8` decode possible and correct.

The telemetry this PR set out to add is live in v1.18.0-beta.21, as a superset of it: the two values you targeted plus target power and target temperature. So #308 is superseded, not wrong in intent. Do you want to retire it yourself now, or keep it open until you have looked at beta.21? Either is fine by me.

The controls work is unchanged and still wanted, as the agreed second step: your write path, the heating rod parameter set command, cmd_func 212 / cmd_id 99. Same terms as on 23 August, recordings of the app writing rather than reconstructions, and the device has to report the setting back so a control cannot lie about its own state.

shuette42:
Two corrections to what I wrote above, both mine.

The capture I called a second owner's PowerOcean is yours. Same config entry, same unit, the download you attached to #247 on 24 August. And I dated your #7 file to the 24th, when it went up on the 23rd with the frames recorded on the 22nd. So the seven `212/33` occurrences and the whole field map come from two downloads off one system, not from two systems. That narrows the evidence base rather than the conclusion, and it was not news either way: I had already identified `212/8` on that #247 capture as the heating rod myself, in that thread, on the day it arrived.

What the second download still does is put the whole watt reading on a second magnitude, 789 W beside 2159 W. That is why it is in the test suite, and it is the only thing I should have claimed for it.

shuette42:
Saw you retired it yourself, thank you.

The credit for it is on main rather than in a promise. The changelog entry now names what your work actually carried: that the reports arrive unsolicited on the connection that is already open, the field names from the excerpt you sanitised by hand, and the download every frame of the field map was read from.

The controls are the piece genuinely still open, and they are yours. Heating rod parameter set, cmd_func 212 / cmd_id 99, which you have already run against your own hardware. Terms unchanged from 23 August: recordings of the app writing rather than reconstructions, and the device has to report the setting back so a control cannot lie about its own state.

Xygen:
Thank you @shuette42 for implementing the complete PowerGlow telemetry support.
I have installed beta.21 and verified that all four entities are now available and reporting values correctly. This is a much more complete solution and fully achieves the original sensor-integration goal. Looking forward to be able to utilize the heating rod as soon as it get's a little colder.

Do you think we should enable the sensor by default at some point?

Thank you also for incorporating the protocol findings and handling the implementation within the existing PowerOcean architecture.

PR #380, opened 2026-09-08
Give every device type its own protobuf command table

shuette42:
## What this changes

The protobuf command registry was one flat `dict[(cmd_func, cmd_id), CmdConfig]` shared by every device. Two device families that mean different messages by the same tuple were kept apart only by the routing order in the coordinator and by a warning in a docstring, neither of which a test can check.

It is now `dict[device_type, dict[(cmd_func, cmd_id), CmdConfig]]`, keyed on the `DEVICE_TYPE_*` constant the classifier already derives from the serial prefix. `decode_proto_runtime_headers` and `decode_proto_runtime_frame` take `device_type` as a required keyword argument, so a caller cannot fall back to a global table by forgetting. Twelve tuples belong to the PowerOcean table, three to the Delta 3 table, and no `CmdConfig` changed.

Design and the alternatives that were rejected are in ADR-024. This is the binding first condition of ADR-008, ahead of giving the PowerPulse 2 its own device type in a later change.

## Why now

The collision is prospective rather than live: neither of the two tuples ADR-008 names is registered today, so nothing on file is mis-decoded. What exists is a shape. A new parser adding a tuple another family already owns produces a silently wrong decode with no test failing, and the next change is exactly that kind of parser.

## Evidence

Behaviour neutrality is measured, not asserted. Every one of the 1764 frames in the 52 recordings on file was replayed before and after: 678 of them decode, to 1307 headers, with zero differences in parse path or key set. The replay tool was itself checked against known answers first: deleting one tuple moves 53 frames out and changes 59, and registering one tuple in a second table raises the two-family counter from 0 to 112.

- 3434 tests pass, 1 skipped
- `ruff check`, `ruff format --check`, and both mypy runs clean

## Review findings

All fixed in the branch. Three local probe scripts called the decoder without the new keyword and would have raised on their first frame, which no gate would have reported since the lint and type gates read the integration and the tests only. Four tests asserted that something did not decode and would have held with their whole table deleted; each now decodes a frame that must hit the table first. The device type without a table of its own had no test at all and now has three, including the header list surviving, which is what the SmartPlug and Delta fallback rests on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.21.0] - 2026-09-08
+
+### Changed
+
+- (pending)
+
 ## [1.20.0] - 2026-09-07
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-- (pending)
+- Nothing about the software changes for anyone using it. Internally, the table that decides which protobuf message a device's frame is read as now exists once per device type instead of once for the whole integration. Until now a single table answered for every device, and two device families that use the same command numbers for different messages were kept apart only by the order in which the code happens to ask, which no test could check. Each family now has its own table and the caller has to say which device it is decoding for, so one family can never read another family's message by accident. Every frame in the recordings on file was replayed before and after the change and produced exactly the same readings.
 
 ## [1.20.0] - 2026-09-07
 
--- a/custom_components/ecoflow_energy/coordinator/mqtt_ingest.py
+++ b/custom_components/ecoflow_energy/coordinator/mqtt_ingest.py
@@ -582,11 +582,13 @@ class MqttIngestMixin(_Base):
 
         if b"\x0a" in payload[:4]:
             try:
-                # Device-type routing comes first: (cmd_func, cmd_id) pairs are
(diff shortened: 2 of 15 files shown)


Xygen:
There are ADRs?
I'd be interested in those. :)

shuette42:
@Xygen Yes, and the register is now in the repository: https://github.com/shuette42/ecoflow-energy-ha/blob/main/documentation/architecture/decisions.md

It holds 23 entries, ADR-002 through ADR-026, oldest first, and each one carries the context at the time, the decision itself, the trade-offs, the alternatives that were rejected and why, and what it changed in the code. Numbers are never reused and never renumbered, so any ADR-NNN you find cited in a comment or a test resolves to exactly one entry. The two this pull request leans on are ADR-024, "The protobuf command registry gets one table per device type", and ADR-008, "PowerPulse 2 (C376) is its own device", so those are the ones to start with. A few numbers are not in the register because they concern how the repository is worked on rather than how the integration behaves, and the intro says so, so a cited number can always be placed.

Your question is the reason it is public. It had been cited by number in comments and tests against a file nobody outside could open. If a decision reads wrong to you or raises a question, an issue or a comment on the file is welcome, and there is a short README beside it on how to read the register: https://github.com/shuette42/ecoflow-energy-ha/blob/main/documentation/architecture/README.md

Xygen:
Great! Thank you so much for your openness and transparency!
That really helps me understand your decisions better and learn from your reasoning.

PR #421, opened 2026-09-16
Add support for the EcoFlow Ocean 2 (RE11, RE17)

### LINKED ISSUE #145: Add support for EcoFlow Ocean 2 (serial prefix RE11)
### Device
EcoFlow Ocean 2, serial prefix `RE11`.

### Current behavior
The integration discovers the device, but the HTTP quota call returns
error 1006 ("current device is not allowed to get device info"). No usable
values in Standard Mode.

### Analysis
Error 1006 means the device is not exposed through the IoT Developer API for
this account, or not bound to the API key. This matches the pattern seen with
the PowerOcean J32D/J32E variants, which are only reachable through Enhanced
Mode. Ocean 2 (RE11) is a new model with no parser or entity definitions yet,
so even with data access it needs a dedicated device profile.

### Next steps
- [ ] Confirm Developer Portal and EcoFlow App use the same account email, verify binding
- [ ] If 1006 persists: capture a data sample via Enhanced Mode (App login)
- [ ] Build parser + entity definitions from the sample
- [ ] Tests + review

### Status
Reporter has the hardware and is willing to test. Tracking here while we
work out data access.

jensfr1:
Ref #145 - the issue stays open until the reporter confirms on a released build.

The Ocean 2 was skipped as an unsupported device with no route to changing that: the Developer API answers error 1006, so there is no polled data to build on, and its live telemetry arrives in a message this integration had never decoded.

**It is not a PowerOcean.** It carries its own frame on `cmd_func` 254 / `cmd_id` 39 and shares no field layout with the `HJ31`/`J32x` line, so routing the prefix to the existing parser — what settled `HJ35`, `J32B` and `P231` — would have created a device whose readings could never fill. It gets its own parser and entity set instead, and lands in `ENHANCED_ONLY_DEVICE_TYPES` so a Developer Key entry is refused rather than set up and left empty.

**`RE11` and `RE17` are one read path**, per your finding in #145: same message families, same payload shape, separated in EcoFlow's device list by power rating and nothing else.

Unlike the Stream and Delta 3 frames the payload is **nested** — readings sit inside submessages rather than at the top level, which a flat field map cannot express. Hence the small generic decoder in `ocean2_proto.py`.

### Scope

Main telemetry frame only, read-only. The per-module battery frame (`cmd_id` 46) carrying cell voltages, temperatures and cycle counts needs its own entity set and follows separately. No write frame from an Ocean 2 has been observed, so there are no controls.

20 sensors: state of charge and remaining energy, signed battery and grid power, total and per-string solar, the house load as the device reports it, the four directional splits, and six counters for the Energy Dashboard.

### Two assignments worth naming

Both look right for a long time, which is why they are in the code comments as well:

- **The field that appears to be the grid meter is the configured feed-in limit.** A constant `10000` on a unit capped at 10 kW, a constant `0` on a zero-export unit — where it is indistinguishable from a real meter at rest. No sensor is built on it, per your note in #145 that it reports the owner's own setting back at them. Grid power comes from the inverter block.
- **The house load is reported in two blocks** that differ by about one measurement cycle. The one the app agrees with wins.

Both were confirmed during a 3.6 kW wallbox charge, which moved all four power paths at once.

### On the readings you flagged

The three battery-block corrections and the additional energy-stream readings are in the module block and the parts of the flow block this PR does not map, so they do not affect what is here. The feed-in ceiling is left out as you asked. Field numbers for the extra energy-stream readings are not in this PR: I have names and types but no frame of my own that pins the numbers, and guessing them would put readings into the dashboard that nothing checked.

### Verification

Full suite green (2408 passed). Two pre-existing failures in `test_frame_capture.py` are unrelated — they need a pytest plugin that does not load on Windows and fail the same way on an unmodified `main`.

The mapping comes from [jensfr1/ha-ecoflow-ocean2](https://github.com/jensfr1/ha-ecoflow-ocean2) and matches the second-installation recording you described in #145.


shuette42:
This is the PR that lands, with your authorship, as said on #145. My own branch follows as a second PR on top of yours once this one is merged: the fixture, the tests around it, and the module frame.

**Fixture result.** I ran your parser against 30 frames cut from the seven-hour second-installation recording named on #145. The sign of `batt_w` from block 87.4 and 7.4 (charge positive) is right on captures 0, 36 and 46. The precedence of block 87 over block 7 for `home_w` is right on captures 35 and 36. State of charge, the three PV strings, and the control replies on 254/40 and 254/42 (ignored) all come out exactly as you mapped them. Your field map holds on hardware you have never had access to.

**Before merge**, in order of effort:

1. Eight tests in `tests/ha/test_init.py` fail on your branch. They use `RE11ZZ1234500001` as the example of a device without a parser, and now RE11 has one (8 failed on your branch, 12 passed on `main`). Fix: change the dummy serial to `ZZ01ZZ1234500001` everywhere it appears as this example, line 1270 (`TestUnroutedDeviceProbeWiring`) and line 1615 (`TestRawCaptureWithoutASkippedDevice`).
2. `_num` in `ocean2_proto.py` lets a non-finite float through. One reaching `home_w` raises inside the precision-0 rounding in `sensor.py` and aborts the rest of that update. The five sibling parsers drop those at the parser. Fix: `_num` returns `None` unless the value is finite.
3. The CHANGELOG entry sits under `[1.22.0] - Unreleased`. A new device model gets its own `[1.23.0] - Unreleased` block above it. `manifest.json` stays at 1.22.0, which is already what you have.
4. The PR body opens with a closing keyword and the issue number. A squash merge would close the issue on the spot, and issues here close only after the reporter confirms on a released build. The commit body already says `Ref #145`, which is right. Please edit the PR body to match.
5. `_pdata_candidates` and `_read_varint` in `ocean2_proto.py` are verbatim copies of the same functions in `stream_proto.py`. Four sibling parsers import them from there. Please do the same and drop the copies.
6. Three small documentation edits: the entity index in `documentation/README.md` has no Ocean 2 line, the Enhanced-only serial list at README line 198 is missing RE11/RE17, and the test dummy serial `RE11ZQH4SF000000` reads like a real fragment where the convention is a clearly fake one such as `RE11TEST00000001`.
7. RE17 is routed from EcoFlow's device list alone, no RE17 frame exists anywhere yet. One sentence in `documentation/entities/ocean-2.md` saying so.

You can apply these yourself, or say the word and I push them onto your branch (maintainer edits are enabled on the PR).

**Two questions.**

Battery power sign. `65.20` is read as an absolute battery power, but in every one of the 13 non-zero fixture frames that also carry block 87.4, it equals exactly `-(87.4)`: 4283 against -4290, -4999 against 4990, -1859 against 1860. The fallback to 65.20 only fires when it is zero, so in the two captures where only block 65 is present (63 and 68), `batt_w` freezes at the last flow value while the pack is charging at up to 4999 W. Can you check the sign of that field once against the app on your unit before this ships?

Solar and grid source. `solar_w` and `grid_w` each have two sources per frame: the unrounded fields 65.4 and 4.13, and the block 87 snapshot. Right now the entity alternates between them depending on which block a frame carries. Across the fixture the block 87 values balance to the same instant (PV minus battery minus grid equals load). 65.4 differs from block 87 by up to 17 W (capture 46: 6687 vs 6670), 4.13 by up to 6 W (capture 36: -373.8 vs -380). My leaning is block 87 for all four power readings, with the unrounded fields only as fallback when block 87 is absent. What does the app show for solar and grid power on your unit? That settles it.

**On my side.** The generic nested decoder duplicates a pattern my own `stream_ac5000_proto._walk` already implements. That is my helper to fix, not something to hand you mid-review, so I will take it. A yes from you and I push it onto your branch after the points above are in. When the CHANGELOG entry moves into the 1.23.0 block, your name goes in it for the mapping, as promised on #145.

Thanks for the PR and for the mapping work behind it. What comes next on the Ocean 2 is the module frame (`cmd_id` 46) and whatever RE41 turns into once a capture from that unit shows up. I would like you on board for that, and for whatever you want to pick up after it.

shuette42:
Pushed one commit (`07695ea`) to fix the CI failure directly rather than wait: `tests/ha/test_init.py` used `RE11ZZ1234500001` as the example of a device with no parser (lines 1270 and 1615), and now that RE11 has one, both tests tried a real login. Changed the dummy serial to `ZZ01ZZ1234500001` in both spots. All 12 tests in those two classes pass locally now; CI is re-running.

Nothing else changed. The rest of the review (the sign question, the source question, the CHANGELOG move, the duplicate helpers) is still yours to take or hand back - happy to push those too once you say which way you want to go.

jensfr1:
Thanks for the fixture run and for pushing the CI fix - that saved a round trip.

Points 2 to 7 are in (`5668fc9`), and the PR body now says `Ref #145` rather than the closing keyword. Locally green on `ruff check`, `ruff format --check`, `mypy` and 2411 tests; the two `test_frame_capture.py` failures are `signal.alarm` on Windows and fail the same way on an unmodified `main`.

**Battery power sign - you are right, and I measured it.**

My unit was charging this morning, so I pulled the raw `65.20` alongside the flow block over 17 consecutive frames:

```
65.20 = -871      87.4 = +880
65.20 = -862      87.4 = +830
65.20 = -852      87.4 = +850
65.20 = -827      87.4 = +830
```

Signed, opposite convention, negative while charging. The magnitudes track each other within the usual inter-block lag rather than matching exactly, which is what I would expect from two blocks with their own measurement instants - but the sign is never in doubt. That is the same thing your 13 fixture frames show, now on a second unit.

So `65.20` is read as `-(65.20)` when the flow block is absent, and the flow block still wins where both are present, since it is the one that balances with the other three readings of its own instant. That closes the freeze you found on captures 63 and 68.

**Solar and grid source - I agree with your leaning, and I would like one more day on the app check.**

Everything on my side points the same way as your fixture. When I compared the two sources on my own installation in August, the block 87 readings balanced and the mixture did not: PV minus battery minus grid landed within a few watts of the reported load in a quiet moment, and hundreds of watts off during a load step, purely because the parts came from different instants. A dashboard built on the mixture shows a house load that does not add up, which is what sent me looking in the first place.

What I have not done is read solar and grid off the app at the same second as a frame, and that is exactly what your question asks. The difference is small enough (17 W, 6 W) that I want to do it properly rather than report an impression. Give me a day.

If you would rather not wait: take block 87 for all four with the unrounded fields as fallback. I do not expect the app check to contradict it, and if it does, it is one line.

**Yes to the nested decoder.** Please push it onto the branch - a helper that already exists in `stream_ac5000_proto` should not be duplicated here, and it is your code to shape.

Happy to stay on for the module frame afterwards. `254/46` is mapped in my own integration already, including the corrections you confirmed on #145 - remaining energy in watt hours, state of health separate from state of charge, and the cell voltage arriving as volts - so that one starts from something rather than from scratch.

jensfr1:
A partial answer on the solar and grid source, and one finding that changes how I read your numbers.

**Block 87 is rounded to 10 W. Block 65 is not.**

Five frames from my unit carrying both fields, within two minutes:

```
65.4 = 1330.0    87.3 = 1330    diff   0 W
65.4 = 1320.0    87.3 = 1320    diff   0 W
65.4 = 1333.0    87.3 = 1330    diff  +3 W
65.4 = 1330.0    87.3 = 1330    diff   0 W
65.4 = 1330.0    87.3 = 1320    diff +10 W
```

Every 87.3 value lands on a multiple of ten; 65.4 does not. So the 17 W you measured on capture 46 and the 6 W on capture 36 are most likely quantisation rather than two different measurement instants. That makes the choice cheaper than it looked: taking block 87 for all four costs up to 10 W of resolution on a reading, and buys four values that belong to one instant and balance against each other. I would make that trade.

**The app check is inconclusive so far, and I want to say why rather than report an impression.**

The portal switches units by magnitude. Above a kilowatt it shows two decimals in kW, so 10 W steps - which cannot separate a rounded 1320 from an unrounded 1325. It showed `1.32 kW` for solar while my frames ran 1320 to 1333, which is consistent with either source.

Below a kilowatt it shows plain watts. That is where the question is decidable, so I will take the reading this evening when the array drops under 1 kW, with a frame captured at the same second, and post both.

If you would rather not hold the PR for that: go with block 87 for all four, fallback to the unrounded fields when it is absent. The rounding finding and the balance argument both point there, and the app reading can only confirm or cost one line.

**On the battery sign from my earlier comment** - that one is settled, 17 frames, no ambiguity, and it is already in `5668fc9`.

shuette42:
Pushed `5cf7ff7` onto your branch, taking the "if you'd rather not wait" option.

Solar and grid power now prefer the balanced flow-block reading (block 87, falling back to block 7), the same way home load and battery power already did. 65.4 and 4.13 are read only as a fallback for a frame that lacks a flow block. No need to hold this on the app check. If it comes back showing something different later, it is a one-line change either way.

Also folded the nested decoder into the declared-path walker in `stream_ac5000_proto.py`, as agreed, so this parser no longer carries a second copy of that logic.

Tests are green: 32 in the Ocean 2 file, 3953 across the suite, `ruff` and `mypy` clean. CI is green too now (tests, lint, HACS validation, hassfest, types). The CHANGELOG credits you by name for the field mapping, the battery-power sign and the solar/grid precedence call, all three confirmed against your own hardware.

Two things I want to say plainly. You fixed six review points and the PR body in a day, and instead of taking the sign question as a note you went and read 17 frames off a charging pack overnight and came back with the raw numbers next to the flow block. That is a better answer than the one I asked for, and it is why this merges with your name on it rather than as evidence for a rewrite. And yes to the module frame - if `254/46` is already mapped on your side with the watt-hour and state-of-health corrections in it, that is the natural next one for you to lead.

And one thing beyond this PR. The last two days are exactly the kind of working relationship I'd want to keep going, not a one-time handoff. If you're up for staying involved past the module frame too, that's what actually gets this integration to the point where it's as solid and complete as we can make it for everyone running it.

PR #425, opened 2026-09-17
Add the Ocean 2 per-module battery frame (254/46)

jensfr1:
Ref #145. The follow-up you offered to hand me — thank you for that.

Twelve readings per module: charge, power, remaining energy, state of health, cycles, three cell temperatures, the power-electronics temperature, module and cell voltage, and current.

### Created on report, not declared for a fixed count

The number of modules is an installation choice — two on my unit, thirteen and fourteen in the recordings you described. A fixed block of sixteen disabled definitions would leave most owners with dozens of entities their hardware can never fill, and the rest hunting through a disabled list for the modules they do have. So every module entity is an `accessory`: sixteen definitions exist as headroom, none costs anything until a module with that number actually reports.

The documentation section follows the `Pack N` template you already use for the PowerOcean packs, registered in `SKIP_SECTIONS`, `TEMPLATE_FAMILIES` and `EXCLUDE_FROM_UNDOCUMENTED` so the per-instance names stay checked against `const.py` rather than merely excluded.

### The three corrections you confirmed in September

Each was the wrong quantity rather than the wrong scale, which is why each looks plausible until it is checked against a moving system:

- **Capacity is the energy remaining.** 4114 Wh at 81.5 %, 4137 Wh at 82.0 %, putting the full module near 5046 Wh. Read as a capacity it falls while the module discharges — harmless on a desk, permanently misleading in service.
- **The state of charge field is the state of health.** Constant at 100.0 across a whole measurement while its neighbour moved.
- **The cell voltage is a cell voltage, not the pack.** It follows the load, and the divide-by-ten it used to carry was the warning sign: a float needs no scaling.

### Two things worth reading before the review

**The module voltage is about 16.5 V**, low enough for a home battery to look wrong. The modules are wired 5S: 16.46 V over 3.311 V per cell is exactly five in series. Confirmed through the power balance — voltage times current hits the reported module power within 1 % on two modules independently.

**The temperatures were separated by a load test, not by their averages.** 45 minutes of wallbox charging at up to 3.6 kW per module. Three readings hold `min <= average <= max` in every frame and rise together and slowly, which is the cells. Four others follow the load within a minute, swinging 11 to 17 K at up to 7 K per minute, which is power electronics. Those four sit on one board and are published as a single maximum: four near-identical entities per module would be noise at fourteen modules. If you would rather have all four, say so — it is a one-line change.

**The module SoC carries no battery device class**, per the singleton rule your `test_const.py` enforces. The system state of charge is the one battery figure for this device.

### Verification

`ruff check`, `ruff format --check`, `mypy` and 2437 tests green. The two failures in `test_frame_capture.py` are `signal.alarm` on Windows and fail the same way on an unmodified `main`.

Seven new parser tests: one module decoded, two modules kept apart in one bundle, the hottest board reading, the three cell temperatures kept separate, a module without a number dropped, out-of-range numbers ignored, and telemetry plus modules read from one payload without either swallowing the other.

What I cannot test here is a bundle of thirteen or fourteen — my unit has two. If your recording still has one of those frames, that is the case I would most like to see run against this.


shuette42:
Ref #145. This is exactly the follow-up I hoped for, and a day after the last one merged.

**On the thirteen and fourteen.** Your docstring reads them as module counts from other installations, and that part is worth correcting - the design around it is not. I went back through the RE11 recording from 2026-09-07, 98 records of `254/46`. Thirteen or fourteen is how many `254/46` headers are bundled into a single frame, not how many modules reported. Across the whole capture there are only two distinct module numbers: field 15 is 1 in 50 records and 2 in 48. What separates the records inside one bundle is field 37, a timestamp stepping 3 to 7 seconds between them. It is a backlog of per-module heartbeats two modules deep, not a snapshot of fourteen modules.

Nothing in the code changes because of that. Keying on the module number you actually see is right whichever way the counts go, and it is what makes the bundle case work at all. It is only the account in the comment that needs the correction. One more thing from the same capture that backs your choice: field 16, the natural module serial, is masked in all 98 records, so field 15 is the only key available. `5.15` to `_index` is the right call.

**One thing to check against your own unit.** The docstring says 39 is the state of health, not the state of charge, and that the pair 38/39 mirrors 2/3 on the older generation. But `_MODULE_FIELD_MAP` declares `"5.3": ("soh_pct", _TYPE_FLOAT, 1)` and never declares 39 at all. In my capture both read a constant 100 across all 98 records, so the data cannot tell me which one you meant. The wire types differ though: field 3 arrives as a varint there, field 39 as a float, and your entry reads 3 as a float. Worth putting the raw frames from your unit next to it before I read the map as final. If your hardware says 3, then my capture and yours disagree about the wire type and that is its own question.

**On the bundle you asked for.** I have that 2026-09-07 recording here with 30 frames of it and will run it against your parser as part of the review. There is also a newer RE41 capture from another owner that likely deserves the same pass once I am into it.

Review starts now. The plan is a v1.23.0-beta.1 within the next few days carrying what #421 shipped, and this one following shortly after in a beta.2 once it is through.

jensfr1:
Your question found two real errors. Both are fixed in `2039a6d`, and both were hidden by my own tests, which built the frames themselves.

**The state of health is 39, read as a float.** The map declared `5.3` — and as a float. On raw frames from my unit field 3 is a varint carrying the same 100, exactly as in your capture, so the walker rejected `5.3` on type and the entity stayed empty. Our hardware does not disagree about the wire types; the map was simply wrong, and the docstring was describing the intent rather than the code. Good catch.

**And while checking that, a second one:** the cell voltage is millivolts. `5.6` reads 3437 on a real frame, and the entity declared volts — it would have published 3437 V. Now published as mV, like your PowerOcean cell-voltage entities. The cross-check is the 5S wiring: five cells at 3.437 V make 17.19 V against the 17.11 V the module reports as its own voltage in field 9.

**The bundle comment is corrected to what your capture shows.** Backlog of per-module heartbeats, timestamp in 37 stepping between the records, bundle size not the module count. Thank you for going back through 98 records to establish that rather than letting a plausible-sounding comment stand.

**On field 16:** it is not masked on my unit — it carries a module serial in plain text. Your capture had been anonymised before it reached you, so that particular conclusion rests on the anonymiser rather than the device. The decision it supports is unchanged: `5.15` is the better key regardless, because a serial has no ordering and would put the module number in a string.

**Two real payloads are now fixtures**, so the next mistake of this kind fails a test instead of shipping. Serials in them are replaced length-preservingly — protobuf carries a length before every field, and a shorter replacement would make everything after it unreadable, which would defeat the point of having a real frame in the first place.

`ruff`, `format`, `mypy` and 2441 tests green.

One thing I would still like from your side when you run the bundle: my fixtures are a resting pack, 0 W and everything thermally flat. The temperatures separate cleanly under load, which is how they were assigned in the first place, but a bundle taken while the pack works would exercise `mos_temp_c` against the cell readings properly. No hurry — the beta plan sounds right, and this one can follow whenever the review is done.

shuette42:
Ref #145. Ran three review lenses against this branch, including your own parser against Oliver's 2026-09-07 RE11 capture (98 records, the same one #421's fixture came from). Pushed `c807726` with what turned out to be mechanical fixes; two things below are yours to decide.

**Confirmed correct against real hardware.** Your parser produces exactly two modules from Oliver's capture, matching the true module count, with every reading in the right physical range: SoC 59.93-99.88 / 60.35-99.81, remaining energy up to 5024 Wh (matches your "near 5046 Wh" full-pack estimate), pack voltage 15.88-17.26 V, cell voltage 3225-3468 mV. The 38/39 fix from your last commit checks out on the wire too - field 3 is a varint in all 98 records, 38 and 39 are both floats, 39 constant at 100.

**Fixed, mechanical:**
- `_parse_module` didn't drop non-finite readings the way `_parse_telemetry` does for its own. Not reachable in this capture (zero NaN/inf anywhere), but a NaN in the module index would have raised `ValueError` at the `int()` cast instead of returning `{}`, and a NaN among the four power-electronics temperatures would have poisoned `max()` and blanked all four instead of just the bad one. Same guard `_parse_telemetry` already carries, added to `_parse_module`, three new tests.
- The twelve module sensor defs were missing `enhanced_only=True` - every other Ocean 2 def has it, and it's the established pattern for an Enhanced-only accessory (the PowerPulse wallbox block does the same). No behavior change today since the accessory gate already blocks creation, but it closes the gap.
- "Fourteen in a recording from another system" reads as fourteen simultaneous modules in the CHANGELOG, README, entity doc and your own docstring. What's actually in the capture is bundles of up to fourteen `254/46` headers in one frame - a heartbeat backlog, two modules deep, the same finding I ran by you on the read-path PR. Your design (create on report) was already right regardless of what the count means; only the prose was describing it wrong. Fixed in all four places.
- A comment block in `const.py` that landed under the wrong import after your last push, moved back.

**Yours to decide, not touched:**
1. **"Module N Max Cell Voltage."** `5 x cell_mv / 1000 - pack_v` comes out negative in 9 of the 98 real records (down to -0.075 V), which a true maximum can't do. Either the label should drop "Max" (it's reading *a* cell, not necessarily the highest one) or there's a reason from the app I'm not seeing - your call, since it's your own hardware to check it against.
2. **Publishing the power-electronics max across a bundle.** Inside one `254/46` record the four-reading max works exactly as designed. Across a bundle, later partial records overwrite earlier full ones through the merge, so a record missing three of the four fields can publish a *lower* max than an earlier record in the same bundle already showed. Zero occurrences in 98 real records (all four fields are always present together here), so this is theoretical on the data we have - flagging it in case a unit with more modules behaves differently, not asking for a fix now.

Docs and translations: 109 doc tests green including two mutation-probe controls, EN/DE key sets identical at 1046 keys each, all 192 module translation keys present and actually translated in German. Nothing to fix there.

Tests: 4006 passed (48 in this file), `ruff`, `ruff format` and `mypy` clean on the pushed commit.

